Issue #5, opened 2019-01-17
setup-debian.yml not able to setup debian but ubuntu instead ..


quabug:
Debian script should be like this depend on [WireGuard doc](https://www.wireguard.com/install/)

```

- name: Setup WireGuard preference
  command: "printf 'Package: *\nPin: release a=unstable\nPin-Priority: 90\n' > /etc/apt/preferences.d/limit-unstable"
  tags:
    - wg-install

- name: Add WireGuard key
  apt_key:
    keyserver: "keyserver.ubuntu.com"
    id: "8B48AD6246925553"
    state: present
  run_once: true
  tags:
    - wg-install

- name: Add WireGuard repository
  apt_repository:
    repo: "deb http://deb.debian.org/debian/ unstable main"
    state: present
    update_cache: yes
  run_once: true
  tags:
    - wg-install
```

githubixx:
Well... The naming is indeed wrong :wink: I'll fix it. But the Ansible Galaxy metadata and also the README states that currently the role only supports Ubuntu 18.04 (Bionic) and Archlinux. Maybe it works also with Ubuntu 16.04 but haven't tested it.
Since I don't use Debian and it's not my primary focus if you're willing to test it I can implement the setup for Debian (or just open a PR). Otherwise I'll just fix the naming and close the issue.

githubixx:
Fixed with https://github.com/githubixx/ansible-role-wireguard/pull/6

PR #6, opened 2019-01-28
feat(debian) enable module to work on debian

ties:
Add support for Debian based on the documentation in debian wiki
and discussion in #5.

Condition for apt cache refresh ("only if Ubuntu") was removed since
the file is only included on Ubuntu.


tasks/setup-debian.yml
```diff
@@ -1,26 +1,28 @@
...
   tags:
     - wg-install
 
 - name: Add WireGuard repository
   apt_repository:
-    repo: "ppa:wireguard/wireguard"
+    repo: "deb http://deb.debian.org/debian/ unstable main"
     state: present
     update_cache: yes
+  run_once: true
```
ties: Copied this from the issue #5. Do not think it is needed. Could fix this and squash if you prefer this

tasks/main.yml
```diff
@@ -3,6 +3,9 @@
   setup:
 
 - include_tasks: "setup-{{ ansible_os_family|lower }}.yml"
+  when: ansible_os_family.lower() == 'archlinux'
```
ties: There are multiple possible values for Archlinux (see https://github.com/cawilliamson/ansible-gpdpocket/blob/master/roles/iso/tasks/main.yml#L3 as an example) so I think this way of splitting between APT and Arch based distributions may make sense...

tasks/setup-debian.yml
```diff
@@ -1,26 +1,37 @@
...
+
+- name: Get architecture
+  shell: dpkg --print-architecture
+  register: dpkg_arch
+  changed_when: "0 != 0"
+
+- name: Install kernel headers to compile wireguard with DKMS
+  apt:
+    name:
+      - "linux-headers-{{ dpkg_arch.stdout }}"
```
ties: There is no good way to do this.

Installign `linux-headers-{{ ansible_kernel }}` would add an explicitly installed package that depends on the currently installed kernel version, which then may not be autoremoved.
Going linux-headers-arch route is not perfect either since the module will not load if the installed kernel is more modern than the running kernel (i.e. scaleways with defaults)

tasks/main.yml
githubixx: At the moment the README and the role metadata clearly states that this role should work on Ubuntu 18.04 and Archlinux. I currently don't have any plans to support Linux Mint & Kali since I can't test this distributions. So I wouldn't care for something that isn't supported anyways :wink:

githubixx: If we want to add Debian support then it would be ok for me to change this `include_tasks` a little bit e.g.:

```
- include_tasks: "setup-{{ ansible_distribution|lower }}.yml"
```
and then have files called `setup-(ubuntu|debian|archlinux).yml`. That should be good enough for now IMHO.

tasks/setup-debian.yml
ties: Indeed it is hard to solve. I have also found an ansible role that alledgedly sets up the kernel headers for scaleways.

I propose that the module should work on "normal" debian derivatives and just fails on exceptions when the installed kernel does not match the kernel that is running.

In addition the ubuntu code may need `linux-headers-generic` package. Could add that easily and I'm ok with testing the ubuntu derivation.

githubixx: I guess this needs to be done completely different. I mean you can install `linux-image-virtual`, `linux-image-generic`, `linux-image-virtual`, `linux-image-generic-hwe`, and what not :wink: Maybe we need to allow to specify the linux kernel image name as a variable. At least optional. As a default fallback the current implementation seems to be good enough for me but allowing the user to override this.

Additionally I was thinking of a cronjob/systemd.timer that periodically checks if the installed kernel and kernel-header are matching and if not installs the kernel-header. That would at least allow to compile the WireGuard dkms.

Now the question is if a new kernel-header was installed should the job also trigger a WireGuard dkms rebuild to ensure that the new kernel has the correct WireGuard module during next reboot? Because ATM that's a problem. The role installs the kernel-header matching the currently in use kernel. And 5 minutes later a new kernel will be auto-installed but that doesn't mean it will also install the kernel-header... So if you now reboot... Tada... No WireGuard. But I also don't want to use ` apt-mark hold ...` to hold the current kernel. Nobody would expect that a WireGuard role also changes your kernel handling. I'm not really sure how to solve this problem.

tasks/main.yml
githubixx: `ansible_distribution == 'archlinux'` isn't needed BTW. The task

- include_tasks: ""setup-{{ ansible_distribution|lower }}.yml""

is sufficient. You just need the include files accordingly e.g.:

```
tasks/setup-archlinux.yml
tasks/setup-ubuntu.yml
tasks/setup-debian.yml
```
It's basically the same as it is now just with the difference that `setup-debian.yml` will then really install the setup for Debian and not Ubuntu :wink: My fault was to use `ansible_os_family` variable instead of `ansible_distribution`. Because the last one will correctly print the distribution and not the OS family which is the same for Debian and Ubuntu of course.

ties: Applied this suggestions. It's cleaner and for arch derivates a user could always override ansible_distribution in inventory :).

tasks/setup-debian.yml
ties: I guess this is a hard problem;

For Debian is is relatively easy. I went for the approach using `dpkg` to get the architecture since I think (did not verify) that `linux-image-amd64` would track the currently installed kernel through its dependencies.

Ubuntu is harder. I had not thought about those variations.

A default per distribution would be featible using [os-dependent-variables](https://ansible-tips-and-tricks.readthedocs.io/en/latest/os-dependent-tasks/variables/).

It is easy to get the currently running kernel name (`ansible_kernel`). Perhaps that could be installed by default on Ubuntu to match the derivation? However this would open up the problem you describe and may prevent kernels from being autoremoved.

The problem with a DKMS rebuild without a kernel version installed almost gives me the feeling that DKMS is broken by design when installed in this way 😟. I have not experienced this failure mode - but my machines likely contained `linux-image-{{ derivation }}`. And I have stayed away from scaleways…

tasks/setup-debian.yml
```diff
@@ -1,26 +1,37 @@
...
+    repo: "deb http://deb.debian.org/debian/ unstable main"
     state: present
     update_cache: yes
   tags:
     - wg-install
+
+- name: Get architecture
+  shell: dpkg --print-architecture
+  register: dpkg_arch
+  changed_when: "0 != 0"
```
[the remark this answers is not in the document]
ties: I copied this verbatim the first time I used it. Without a `changed_when` argument the shell module would indicate that something had changed in this play.

After [looking it up](https://docs.ansible.com/ansible/latest/user_guide/playbooks_error_handling.html#overriding-the-changed-result) it seems like `changed: False` is also possible, I will change to this alternative.

tasks/setup-debian.yml
ties: "enabling DKMS modules to be compiled" sounds like a valid ansible module and a good dependency for this one. And for now this module is idempotent 🙂.

ties:
I (just) tested this role locally using a set-up with 3 clean vagrant virtual machines and the VPN came up as expected 🙂.

tasks/setup-debian.yml
```diff
@@ -1,26 +1,37 @@
...
+
+- name: Get architecture
+  shell: dpkg --print-architecture
+  register: dpkg_arch
+  changed_when: False
+
+- name: Install kernel headers to compile wireguard with DKMS
+  apt:
+    name:
+      - "linux-headers-{{ dpkg_arch.stdout }}"
```
ties: There is a funny thing happening here by the way. In the `ansible -m setup localhost` output there is `facter_os.architecture` which contains `amd64`.

In `ansible_architecture` this is normalized to 'x86_64' and not usable.

ties:
I tried to use `facter_os['architecture']` instead of the output of the `dpkg` command the branch runs but can not get it to work unfortunately from the tasks. The (seemingly derived) `ansible_architecture` variable is not usabe (`x86_64` instead of `amd64`).

I do not know if `facter` is always available when running ansible. The `ansible_os` data does not seem to be available on my laptop (Fedora) and on a Ubuntu machine.

tasks/setup-debian.yml
githubixx: I can find this variables:
```
    "ansible_architecture": "x86_64",
    "ansible_machine": "x86_64",
    "ansible_userspace_architecture": "x86_64",
```
But none with `amd64`. But at least it is consistent and AFAIK it's also the official declaration for a x86 64bit CPU.

githubixx:
I'll merge this now. Thx for your contribution!

tasks/setup-debian.yml
ties: x86_64 is the consistent way of describing the architecture. On some machines I see the following - but definitely out of scope for this change. Possibly good for the dkms-headers role:
```js
# ansible -m setup localhost | sed -e 's/.* =>//g' | jq '.ansible_facts.facter_os'
{
  "architecture": "amd64",
  "distro": {
    "codename": "buster",
    "description": "Debian GNU/Linux buster/sid",
    "id": "Debian",
    "release": {
      "full": "testing",
      "major": "testing"
    }
  },
  "family": "Debian",
  "hardware": "x86_64",
  "name": "Debian",
  "release": {
    "full": "buster/sid",
    "major": "buster/sid"
  },
  "selinux": {
    "enabled": false
  }
}
```

PR #9, opened 2019-06-04
add CentOS support

ahanselka:
This simply adds CentOS/RHEL support. I've tried this out on my own network and it appears to work as expected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 ---------
 
+**3.2.0**
+
+- add support for RHEL/CentOS (contribution by ahanselka)
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 Changelog
 ---------
 
+**UNRELEASED**
```
githubixx: I added your previous PR as version `3.1.0`. Please rebase on the new master and instead of `UNRELEASED` the version would be `3.2.0` as we just adding a feature and do not break existing stuff.

ahanselka: Cool, it has been done. Thanks!

PR #13, opened 2019-06-14
fix problem with different ports on hosts

lukasz-aksamit:
This fixes a bug when user has different wireguard ports on hosts.

### DIFF
--- a/templates/wg.conf.j2
+++ b/templates/wg.conf.j2
@@ -31,7 +31,7 @@ SaveConfig = true
     {% if hostvars[host].wireguard_endpoint is not defined %}
     Endpoint = {{host}}:{{wireguard_port}}
     {% elif hostvars[host].wireguard_endpoint != "" %}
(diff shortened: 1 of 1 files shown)


githubixx:
Thanks for the pull request. But this change needs to be extentend. Currently `wireguard_port` is a "global" variable defined in `group_vars/` directory. So every host uses the same port. If you want a per host `wireguard_port` you need to add a additional `if` block. If a host `wireguard_port` is defined use that one if not use the global `wireguard_port`. We can't assume that everybody will define the `wireguard_port` for every host. This would at least break current setups.

dennisse:
But.. If you define `wireguard_port` in `host_vars/`, it just overrides the variable in `group_vars/`. There is no need to specify where the variable should come from.

https://docs.ansible.com/ansible/latest/user_guide/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable

githubixx:
Since I don't have the time currently to implement it myself and the PR wasn't updated since nearly 25 days I guess we can close this for now.

PR #19, opened 2019-09-18
Cleans unecessary files.

pierreozoux:
Not sure you are using this test directory, if you are, let's keep it.
But I like to remove unecessary dirs ;)


githubixx:
Well I don't need it but they're created by `ansible-galaxy init ...`. So I would like to keep them. I'm not sure if this files are needed (at least a skeleton) to be able to upload a role to Ansible Galaxy.

pierreozoux:
No, I don't think they are, let's try.
I have a life philosophie that tells me that each line of code, I have to maintain for the next 10 years ;) Each time I remove lines, I'm happier ;)

PR #20, opened 2019-09-18
Removes unecessary lines.

pierreozoux:
Not sure what is the objective of these lines, but I think they are useless.
Check the doc here to be sure:
https://docs.ansible.com/ansible/latest/plugins/lookup/inventory_hostnames.html

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -67,8 +67,6 @@
 - name: Generate WireGuard private key
   shell: "wg genkey"
   register: wg_private_key_result
-  with_inventory_hostnames:
-    - vpn
   when: not private_key_file_stat.stat.exists
   tags:
     - wg-generate-keys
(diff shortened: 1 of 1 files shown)


githubixx:
Maybe you're right but I'll keep them for now as I don't have time to test it thoroughly at the moment. Normally I don't add something just for fun. But it could be indeed a leftover from a former implementation and I forgot to remove it. But as it also doesn't hurt I keep it for now until I've more time. So we can keep this issue open for now.

githubixx:
Ok I tested it. It's indeed not needed. In fact it was a wired loop that basically worked by "accident" so to say. So thanks for pointing it out. But just removing this line doesn't work. Then the `Set private/public key fact` tasks fail. I implemented a fix. Will upload shortly.

githubixx:
https://github.com/githubixx/ansible-role-wireguard/pull/27 merged. Closing this one then.

PR #22, opened 2019-09-18
Adds molecule tests.

pierreozoux:


molecule/default/playbook.yml
```diff
@@ -0,0 +1,20 @@
...
+      state: present
+    when: hostvars[item]['ansible_enp0s8']['ipv4']['address'] is defined
+    with_items:
+      - "{{ groups['all'] }}"
+
+- name: Test
+  hosts: all
+  become: true
+  roles:
+    - role: ansible-role-wireguard
```
githubixx: This didn't worked for me but `githubixx.wireguard` worked.

README.md
```diff
@@ -252,6 +252,31 @@ Example Playbook
     - wireguard
 ```
 
+Run tests
+---------
+
+Make sure to have vagrant installed.
```
githubixx: Just installing vagrant isn't enough - at least not for Archlinux. It should be also mentioned that this either also needs `VirtualBox` installed (as this is the default here) or the Vagrant libvirt plugin (as described in the Archlinx wiki e.g. https://wiki.archlinux.org/index.php/Vagrant#vagrant-libvirt). At least we should make people aware that not only Vagrant needs to be installed but also a suitable provider like `VirtualBox` or `libvirt` plugin.

molecule/default/playbook.yml
pierreozoux: Hum, this is strange.. I think you have a local configuration that is "special".
What is the name of your folder that holds the role?
From what I understand, molecule will add the local folder to the playbook paths (But I;m not 100% sure).

README.md
pierreozoux: done

pierreozoux:
Yeah, it was not straight forward for me neither ;) But I was happy when I succeeded ;) It is my first molecule test as well ;)

molecule/default/playbook.yml
githubixx: These Ansible Galaxy guys really make me angry. They're changing stuff without further notice the second time now... But I guess we were both wrong.
To install the role via Galaxy `ansible-galaxy install githubixx.ansible_role_wireguard` is needed. So I'm pretty sure that you need
```
  roles:
    - role: githubixx.ansible_role_wireguard
```
here. My local directory was called `githubixx.wireguard` and that was the role name at the time I uploaded the role but now it's called `githubixx.ansible_role_wireguard` for whatever reasons. And if I use `- role: githubixx.ansible_role_wireguard` it works for me. And this is the only value that makes sense IMHO because the official role name in Ansible Galaxy is `githubixx.ansible_role_wireguard`.
I assume that you `git clone ...` the role from Github. In that case `- role: ansible-role-wireguard` makes sense but we should stay with the "official" role name in Galaxy.

molecule/default/molecule.yml
```diff
@@ -0,0 +1,60 @@
...
+        type: static
+        ip: 192.168.11.4
+        auto_config: true
+    groups:
+      - vpn
+
+provisioner:
+  name: ansible
+  inventory:
+    host_vars:
```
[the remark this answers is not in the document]
pierreozoux: To be honest, I just quickly tested it with master, I tested it more with the stateless PR #23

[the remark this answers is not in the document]
githubixx: If it's intended to be used with the stateless PR then I would suggest to close this PR and merge this changes into the stateless PR.

githubixx:
A slightly different approach was implemented with https://github.com/githubixx/ansible-role-wireguard/pull/83

PR #23, opened 2019-09-18
Makes the role stateless.

### LINKED ISSUE #18: Need for storing state locally?
Hi,

thanks for writing and sharing this role, this is really nice!

I have a question, is it necessary to store the state locally?
It would be easy to check if the server is already configured, find the private key and derive the public key, and if not, create the private key?

Am I missing something?

If you agree that storing the state is not so necessary, I can PR, to remove this need.

Cheers!

pierreozoux:
A lot of code removed ;)

closes #18

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -1,16 +1,4 @@
 ---
-# The LOCAL directory where the WireGuard certificates are stored after they
-# were generated. By default this will expand to user's LOCAL ${HOME}
-# (the user that run's "ansible-playbook" command) plus
-# "/wireguard/certs". That means if the user's ${HOME} directory is e.g.
-# "/home/da_user" then "wireguard_cert_directory" will have a value of
-# "/home/da_user/wireguard/certs". If you change this make sure that
-# the parent directory is writable by the user that runs "ansible-playbook"
-# command.
-wireguard_cert_directory: "{{ '~/wireguard/certs' | expanduser }}"
-wireguard_cert_owner: "root"
-wireguard_cert_group: "root"
-
 # Directory to store WireGuard configuration on the remote hosts
 wireguard_remote_directory: "/etc/wireguard"
 
@@ -19,3 +7,6 @@ wireguard_port: "51820"
 
 # The interface name that wireguard should use.
 wireguard_interface: "wg0"
+
+# Wireguard config file path
+wg_conf_path: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -28,46 +28,35 @@
   tags:
     - wg-install
 
-- name: Create WireGuard certificates directory
-  file:
-    dest: "{{ wireguard_cert_directory }}"
-    state: directory
-    owner: "{{ wireguard_cert_owner }}"
-    group: "{{ wireguard_cert_group }}"
-    mode: 0700
-  run_once: true
-  delegate_to: localhost
-  tags:
-    wg-generate-keys
-
 - name: Set WireGuard IP (without mask)
   set_fact:
     wireguard_ip: "{{ wireguard_address.split('/')[0] }}"
 
-- name: Set path to private key file
-  set_fact:
-    private_key_file_path: "{{ wireguard_cert_directory }}/{{ inventory_hostname }}.private.key"
-  tags:
-    wg-generate-keys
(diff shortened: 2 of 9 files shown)


pierreozoux:
Hello!

What do you prefer?

Do you prefer to merge PR one by one? Or do you prefer I create a big PR with all changes in different commits?

I'd like to go further on this topic, but if you can't merge right now, I'll adopt a different strategy.
No stress :) I'd like to avoid the free software burn out ;)

Cheers!

Pierre

githubixx:
Hi! TBH, I have no idea :wink: Normally Smaller changes are way easier to review. But of course it always depends. So I would say make it as small as possible and as big as necessary :smile: I'll try to catch up with the PRs today or tomorrow. But I would also like to develop kind of a upgrade/migration guide for current users if possible. And since I'm a current user I hopefully be able to create one. I guess it will hurt anyways but maybe it's possible to reduce the pain a little bit...

githubixx:
This PR currently has merge conflicts because of https://github.com/githubixx/ansible-role-wireguard/pull/27. But should be easy to adjust.

pierreozoux:
I'll adjust today, thanks for your time :)

pierreozoux:
Ok, adujsted.

githubixx:
I was not sure at the beginning but since this is already a breaking change we should also include the changes to support multiple Wireguard interfaces as discussed in https://github.com/githubixx/ansible-role-wireguard/issues/25 + the Molecule tests for which you already created the PR  https://github.com/githubixx/ansible-role-wireguard/pull/22. Then we've all in one which makes it easier to test. I hope this works for you.

defaults/main.yml
```diff
@@ -1,16 +1,4 @@
...
-# (the user that run's "ansible-playbook" command) plus
-# "/wireguard/certs". That means if the user's ${HOME} directory is e.g.
-# "/home/da_user" then "wireguard_cert_directory" will have a value of
-# "/home/da_user/wireguard/certs". If you change this make sure that
-# the parent directory is writable by the user that runs "ansible-playbook"
-# command.
-wireguard_cert_directory: "{{ '~/wireguard/certs' | expanduser }}"
-wireguard_cert_owner: "root"
-wireguard_cert_group: "root"
-
```
githubixx: This part also needs to be removed from README.md in the `Role Variables` section.

githubixx: Also the CHANGELOG.md needs to be updates. I guess when we ready with this PR we need to adjust it a little bit again but for now I guess we can already add this text:

```
**4.0.0**

- IMPORTANT: Version 4.0.0 has some breaking changes and is no longer compatible with versions < 4.0.0!
- Role is now stateless. It doesn't store private and public keys anymore on your local machine. Therefore the variables `wireguard_cert_directory`, `wireguard_cert_owner` and `wireguard_cert_group` are now obsolete and no longer needed (contribution by pierreozoux)
```
If you merge the code for supporting multiple Wireguard interfaces into this PR then the CHANGELOG text above needs to be extended e.g.:
```
- Role now supports multiple Wireguard interfaces. README was updated accordingly (contribution by pierreozoux)
```
So please also update README accordingly when you implement https://github.com/githubixx/ansible-role-wireguard/issues/25.

tasks/main.yml
```diff
@@ -21,114 +21,68 @@
...
+- name: Check if WireGuard file exists
+  stat:
+    path: "{{ wg_conf_path }}"
+  register: wg_conf
 
-- name: Set path to public key file
-  set_fact:
-    public_key_file_path: "{{ wireguard_cert_directory }}/{{ inventory_hostname }}.public.key"
+- name: Read WireGuard private key
+  shell: "cat {{ wg_conf_path }} | grep PrivateKey | cut -d= -f2- | tr -d [:space:]"
```
githubixx: `sed` should be good enough:
```
shell: "sed -n 's/PrivateKey = \(.*\)/\1/p' {{ wg_conf_path }}"
```

githubixx: Maybe even `command` module may work:
```
- name: Read WireGuard private key
  command:
    argv:
      - sed
      - -n
      - 's/PrivateKey = \(.*\)/\1/p'
      - {{ wg_conf_path }}
```

githubixx:
Functionality implemented with https://github.com/githubixx/ansible-role-wireguard/pull/29 .

Issue #26, opened 2019-09-19
Loop over all hosts from the play


pierreozoux:
There is one more thing I'm not a fan.

https://github.com/githubixx/ansible-role-wireguard/blob/master/templates/wg.conf.j2#L19

I think the role should be agnostic of the groups the role user is using.

I think it would be better to use [ansible_play_hosts](https://docs.ansible.com/ansible/latest/reference_appendices/special_variables.html) and in your playbook, you tell the role to only use the vpn group. Basically it is the responsibility of the playbook to decide which hosts, and not the role.

If you agree, I'll refactor that part too.

githubixx:
Yeah, I was never happy with that too. Makes sense to refactor.

githubixx:
Implemented by https://github.com/githubixx/ansible-role-wireguard/pull/29

Issue #25, opened 2019-09-18
Support for mutliple wireguard vpn on same host.


pierreozoux:
Ok, now I know it is a bit more I'm asking, but I really don't want to fork your project, and I'd prefer to collaborate here :)

Basically, We run on hetzner too, and we have a kubernetes cluster on baremetal instance, and we have ceph, and 10g net cards.
So I'd like one wireguard for kube, one for ceph frontend, and one for ceph backend.

I see how I can do it with your role, but it will not be backward compatible.

Basically:

```
#instance-1 host_var
wireguard:
 - wg0:
    - address: 10.0.0.1/24
       port: 51820
 - ceph-backend:
    - address: 10.0.1.1/24
       port: 51821
#instance-2 host_var
wireguard:
 - wg0:
    - address: 10.0.0.2/24
       port: 51820
 - ceph-backend:
    - address: 10.0.1.2/24
       port: 51821
```

And then I'd loop over these keys and derive the interface name from there.

What do you think? Can I PR? (I'll do it anyway for me ;) )

githubixx:
You give me a hard time, man... :wink: This is a feature I already thought about since I started with that role but never had time to implement it. Well, yeah, then lets do it. Create the PR when you're ready. But we still need to find a way to prevent the execution of the new version of this role for people who already use the current version as they would basically destroy they setup otherwise if they just upgrade without reading the changelog (e.g. by checking the existence of `wireguard_address` variable and stop executing the role if this variable exists). Maybe I can provide some upgrade script...

AnatomicJC:
Hi there, I forked this repo to implement this requirement too, for my personal needs.

I implemented it this way:

```
vpn:
  vars:
    wireguard_ansible_children_group_prefix: "childvpn"
  children:
    childvpn1:
      vars:
        childvpn1_interface: "wg1"
        childvpn1_port: 51821
      hosts:
        k8s-1:
          childvpn1_address: "10.9.0.101/24"
          wireguard_endpoint: "{{ ansible_default_ipv4.address }}"
        k8s-2:
          childvpn1_address: "10.9.0.102/24"
          wireguard_endpoint: "{{ ansible_default_ipv4.address }}"
        k8s-3:
          childvpn1_address: "10.9.0.103/24"
          wireguard_endpoint: "{{ ansible_default_ipv4.address }}"
    childvpn2:
      vars:
        childvpn2_interface: "wg2"
        childvpn2_port: 51822
      hosts:
[...]

Issue #17, opened 2019-07-08
Make directory/file mode bits configurable


githubixx:
Currently the directory specified in `wireguard_cert_directory` has fixed permissions set to `0700`. A `wireguard_cert_directory_permission` variable should be introduced with a default setting of `0700` to make it configurable but keeping the current setting as default.

Also the file permissions of the certificate files should be configurable while using the current settings as defaults to not change the current behaviour.

githubixx:
Not needed anymore because of https://github.com/githubixx/ansible-role-wireguard/pull/29

PR #29, opened 2019-10-18
Merge stateless idea with no local storage of public and private keys, support multiple interface per hosts using several groups

fbourqui:
In addition:
- add reload module on update config file

Solve:
#18
#25
#26

#17, is now irrelevant

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -38,25 +38,34 @@
   tags:
     - wg-generate-keys
 
-- name: Generate WireGuard private key
-  shell: "wg genkey"
-  register: wg_private_key_result
+- block:
+  - name: Generate WireGuard private key
+    shell: "wg genkey"
+    register: wg_private_key_result
+    tags:
+      - wg-generate-keys
+
+  - name: Set private key fact
+    set_fact:
+      private_key: "{{ wg_private_key_result.stdout }}"
+    tags:
+      - wg-generate-keys
   when: not config_file_stat.stat.exists
-  tags:
-    - wg-generate-keys
 
-- name: Read WireGuard config file
-  slurp:
-    src: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
-  register: wg_config
-  tags:
-    - wg-config
+- block:
+  - name: Read WireGuard config file
+    slurp:
+      src: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
+    register: wg_config
+    tags:
+      - wg-config
 
-- name: Set private key fact
-  set_fact:
-    private_key: "{{ wg_config['content'] | b64decode | regex_findall('PrivateKey = (.*)') | first }}"
-  tags:
-    - wg-config
+  - name: Set private key fact
+    set_fact:
+      private_key: "{{ wg_config['content'] | b64decode | regex_findall('PrivateKey = (.*)') | first }}"
+    tags:
+      - wg-config
+  when: config_file_stat.stat.exists
 
 - name: Derive WireGuard public key
   shell: "echo '{{ private_key }}' | wg pubkey"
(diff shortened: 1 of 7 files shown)


fbourqui:
Solve:
#18 
#26

tasks/main.yml
```diff
@@ -28,106 +27,58 @@
...
-- name: Register if private key already exists
+- name: Register if config/private key already exists on target host
   stat:
-    path: "{{ private_key_file_path }}"
-  register: private_key_file_stat
-  delegate_to: localhost
+    path: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
+  register: config_file_stat
   tags:
     - wg-generate-keys
```
laurawadden: I think you have to add the 'wg-config' tag here also.

tasks/main.yml
```diff
@@ -28,106 +27,58 @@
...
-    - wg-generate-keys
-
-- name: Generate WireGuard public key
-  shell: "echo '{{ wg_private_key }}' | wg pubkey"
+- block:
+  - name: Generate WireGuard private key
+    shell: "wg genkey"
+    register: wg_private_key_result
+    tags:
+      - wg-generate-keys
```
laurawadden: I would also add a `when: not config_file_stat.stat.exists` to this task as well as the one below it.

fbourqui: No, it's a block, the wen apply to the full block:
```
- block:
  - name: Generate WireGuard private key
    shell: "wg genkey"
    register: wg_private_key_result
    tags:
      - wg-generate-keys

  - name: Set private key fact
    set_fact:
      private_key: "{{ wg_private_key_result.stdout }}"
    tags:
      - wg-generate-keys
  when: not config_file_stat.stat.exists
```

laurawadden: Got it, thanks! I had accidentally indented my 'when'.

tasks/main.yml
fbourqui: No, we check if the config exist, to know if a key need to be generated, the config generation is done later with the generated key or the key existing in config file.

laurawadden: But if you run `-t wg-config` alone without other tags, I thought that it needs to register whether or not the private key already exists.

fbourqui: If you run -t wg-config, your assuming that the key exist already, as even is the config is not existing you would not generate a new one. So unless you merge tags "wg-config" and "wg-generate-keys", or put "wg-config" in all task with tag "wg-generate-keys

fbourqui: OK, it make sense to add the tag, to not faill this block:

```
- block:
  - name: Read WireGuard config file
    slurp:
      src: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
    register: wg_config
    tags:
      - wg-config

  - name: Set private key fact
    set_fact:
      private_key: "{{ wg_config['content'] | b64decode | regex_findall('PrivateKey = (.*)') | first }}"
    tags:
      - wg-config
  when: config_file_stat.stat.exists
```

fbourqui: Tad was added in latest commit.

tasks/main.yml
```diff
@@ -28,106 +27,59 @@
...
+
+  - name: Set private key fact
+    set_fact:
+      private_key: "{{ wg_config['content'] | b64decode | regex_findall('PrivateKey = (.*)') | first }}"
+    tags:
+      - wg-config
+  when: config_file_stat.stat.exists
+
+- name: Derive WireGuard public key
+  shell: "echo '{{ private_key }}' | wg pubkey"
```
githubixx:
```suggestion
  shell: "echo '{{ private_key }}' | wg pubkey" # noqa 306
```
ansible-lint complains that pipes should set the `pipefail` (bash) option which I would agree in general. But in that case it's not really needed as the last command `wg pubkey` will fail anyways if the `private_key` isn't valid. So by adding `noqa 306` ansible-lint will skip this check in that line.

pierreozoux:
Thanks a lot @fbourqui :)

PR #32, opened 2019-12-26
Add basic fedora support

ties:
This seems to be enough to enable wireguard on my fedora machine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+**4.2.0**
+
+- Add support for Fedora (contribution by @ties)
+
+
 **4.1.1**
 
 - Install GPG to be able to import WireGuard key (Debian)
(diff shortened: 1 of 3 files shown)


githubixx:
Thanks for you contribution! Please add an entry to `CHANGELOG.md`:
```
**4.2.0**
- add support for Fedora (contribution by ties)
```
Also `platforms` in `meta/main.yml` needs to be updated accordingly. A list of supported platforms can be found here: https://github.com/ansible/galaxy/issues/52

ties:
> Also `platforms` in `meta/main.yml` needs to be updated accordingly. A list of supported platforms can be found here: [ansible/galaxy#52](https://github.com/ansible/galaxy/issues/52)

I erred on the side of caution with the supported versions. Have tested this with Fedora 31 and I am _quite_ sure it would work on 30, but have not tested the latter.

Issue #33, opened 2019-12-28
Adding Peers while wg interface is UP


christf:
Adding a peer to the ansible config results in the existing wg interface to be destroyed and restarted.
Instead it would be nice if just the peer was added to the already-existing configuration.

githubixx:
Solved by https://github.com/githubixx/ansible-role-wireguard/pull/38

PR #34, opened 2020-01-05
Allow to add peers to wg interfaces without taking them offline

christf:
Existing connections to peers should not be impaired when adding new peers. This PR implements online config sync for wireguard to allow just that.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -3,3 +3,5 @@
   service:
     name: "wg-quick@{{ wireguard_interface }}"
     state: restarted
+- name: start/reload wireguard
+  command: "bash -c 'systemctl start wg-quick@{{ wireguard_interface }}; wg syncconf {{ wireguard_interface}} <(wg-quick strip /etc/wireguard/{{ wireguard_interface }}.conf)'"
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -99,7 +99,7 @@
   tags:
     - wg-config
   notify:
-    - restart wireguard
+    - start/reload wireguard
 
 - name: Check if reload-module-on-update is set
   stat:


christf:
this closes #33

handlers/main.yml
```diff
@@ -3,3 +3,5 @@
   service:
     name: "wg-quick@{{ wireguard_interface }}"
     state: restarted
+- name: start/reload wireguard
+  command: "bash -c 'systemctl start wg-quick@{{ wireguard_interface }}; wg syncconf {{ wireguard_interface}} <(wg-quick strip /etc/wireguard/{{ wireguard_interface }}.conf)'"
```
githubixx:
```suggestion
  shell: |
      set -o errexit
      systemctl is-active wg-quick@wg-quick@{{ wireguard_interface }} || systemctl start wg-quick@{{ wireguard_interface }}
      wg syncconf {{ wireguard_interface }} <(wg-quick strip /etc/wireguard/{{ wireguard_interface }}.conf)
      exit 0
  args:
    executable: "/usr/bin/env bash"
```
I haven't tested the code above yet. So my suggested change might not work right away. In general it's more readable. And  as you execute `bash` anyways the `shell` module can be used right away together with the `executable` parameter that contains the path to `bash`. I'm just not sure if `/usr/bin/env bash` actually would work for `executable`. In general `/usr/bin/env bash` is more portable but if Ansible doesn't like that value then an absolute path to the `bash` executable is necessary e.g. `/bin/bash`. I guess `/bin/bash` should work for all distributions. Nowadays `/usr/bin/bash` is more common but `/bin/bash` should be still in place at least as a softlink for compatibility reasons. `/bin/bash` exists for Ubuntu 16.04/18.04/19.10 and Debian Stretch. Ubuntu 19.10 additionally has `/usr/bin/bash`. Fedora 31 and Archlinux has `bash` in both paths. So `/bin/bash` should be save.

Additionally an entry in the `CHANGELOG.md` is needed e.g.:
```
**5.0.0**

- Avoid restarting Wireguard interfaces in case of configuration change (contribution by @christf)
```
As this is a potential breaking change increasing the major release version seems to be reasonable to me.

githubixx: Tested a little bit. `executable:` needs the value `"/bin/bash"` as already suspected. Besides that if figured out that `wg syncconf` is pretty new. So only Wireguard releases that came out around November have this subcommand. So the old `restart wireguard` handler is still needed for Wireguard setups with older releases.
To be able to handle both cases I think it makes sense to check if `wg syncconf` works or returns an error and register the result in a variable. Depending on the variable the handler could be set accordingly. Or if the this variable stores the handler name directly it maybe could be used as the value for `notify:` e.g. That are my first thoughts. Maybe there are even better solutions. 
I still don't like all the shell scripting but I also don't want to implement a Ansible module :wink: But avoiding to restart the Wireguard interface after a configuration change makes the shell script acceptable as it adds a lot of value :smiley:

githubixx:
Solved by https://github.com/githubixx/ansible-role-wireguard/pull/38.

PR #35, opened 2020-01-18
Ability to create multiple postup/postdown/preup/predown commands

Madic-:
Often, more than one iptables command is required for a meaningful firewall setup. Of course we can just append all commands with '&&' but that isn't readable depending on the amount of commands.

With this pull request, multiple  postup/postdown/preup/predown commands are created if specified


githubixx:
PR merged. Thanks!

PR #41, opened 2020-02-16
Configure unmanaged hosts

juju4:
- allow to configure unmanaged hosts
- idempotence check with changed_when: false

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,6 +257,19 @@ Endpoint = server.at.home.p.domain.tld:51820
 
 The other WireGuard config files (`wg0.conf` by default) looks similar but of course `[Interface]` includes the config of that specific host and the `[Peer]` entries lists the config of the other hosts.
 
+Following variable allows to configure unmanaged hosts which are not configured by Ansible, typically smartphones and tablets.
+In this case, private key and client configuration are generated on wireguard hub.
+They can be shared through qrcode.
+```
+wireguard_unmanaged_hosts:
+  - host: 'test'
+    allowed_ips: 10.200.10.10/32
+    dns: 10.200.10.1
+  - host: 'test2'
+    allowed_ips: 10.200.10.11/32
+    dns: 10.200.10.1
+```
+
 Example Playbook
 ----------------
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ I used [PeerVPN](https://peervpn.net/) before but that wasn't updated for a whil
 
 In general WireGuard is a network tunnel (VPN) for IPv4 and IPv6 that uses UDP. If you need more information about [WireGuard](https://www.wireguard.io/) you can find a good introduction here: [Installing WireGuard, the Modern VPN](https://research.kudelskisecurity.com/2017/06/07/installing-wireguard-the-modern-vpn/).
 
-This role was tested with Ubuntu 18.04 (Bionic Beaver), Debian 9 (Stretch), Archlinux, Fedora 31 and CentOS. It might also work with Ubuntu 16.04 (Xenial Xerus) and Debian 10 (Buster) but haven't tested it. If someone tested it let me please know if it works ;-)
+This role was tested with Ubuntu 18.04 (Bionic Beaver), Debian 9 (Stretch), Archlinux, Fedora 31 and CentOS. It might also work with Ubuntu 16.04 (Xenial Xerus), Debian 10 (Buster) or other distributions but haven't tested it. If someone tested it let me please know if it works or send a pull request to make it work ;-)
 
 Versions
 --------
@@ -97,7 +97,7 @@ wireguard_postdown:
 wireguard_save_config: "true"
 ```
 
-`wireguard_(preup|predown|postup|postdown)` are specified as lists e.g.:
+`wireguard_(preup|predown|postup|postdown)` are specified as lists. Here are two examples:
 
 ```
 wireguard_postup:
@@ -106,7 +106,13 @@ wireguard_postup:
   - iptables -A FORWARD -o %i -j ACCEPT
 ```
 
-If more `iptables` commands needs to be specified e.g. then a list makes it more readable. The commands are executed in order as described in [wg-quick.8](https://git.zx2c4.com/wireguard-tools/about/src/man/wg-quick.8).
+```
+wireguard_preup:
+  - echo 1 > /proc/sys/net/ipv4/ip_forward
+  - ufw allow 51820/udp
+```
+
+The commands are executed in order as described in [wg-quick.8](https://git.zx2c4.com/wireguard-tools/about/src/man/wg-quick.8).
 
 `wireguard_address` is required as already mentioned. It's the IP of the interface name defined with `wireguard_interface` variable (`wg0` by default). Every host needs a unique VPN IP of course. If you don't set `wireguard_endpoint` the playbook will use the hostname defined in the `vpn` hosts group (the Ansible inventory hostname). If you set `wireguard_endpoint` to `""` (empty string) that peer won't have a endpoint. That means that this host can only access hosts that have a `wireguard_endpoint`. That's useful for clients that don't expose any services to the VPN and only want to access services on other hosts. So if you only define one host with `wireguard_endpoint` set and all other hosts have `wireguard_endpoint` set to `""` (empty string) that basically means you've only clients besides one which in that case is the WireGuard server. The third possibility is to set `wireguard_endpoint` to some hostname. E.g. if you have different hostnames for the private and public DNS of that host and need different DNS entries for that case setting `wireguard_endpoint` becomes handy. Take for example the IP above: `wireguard_address: "10.8.0.101"`. That's a private IP and I've created a DNS entry for that private IP like `host01.i.domain.tld` (`i` for internal in that case). For the public IP I've created a DNS entry like `host01.p.domain.tld` (`p` for public). The `wireguard_endpoint` needs to be a interface that the other members in the `vpn` group can connect to. So in that case I would set `wireguard_endpoint` to `host01.p.domain.tld` because WireGuard normally needs to be able to connect to the public IP of the other host(s).
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ I used [PeerVPN](https://peervpn.net/) before but that wasn't updated for a whil
 
 In general WireGuard is a network tunnel (VPN) for IPv4 and IPv6 that uses UDP. If you need more information about [WireGuard](https://www.wireguard.io/) you can find a good introduction here: [Installing WireGuard, the Modern VPN](https://research.kudelskisecurity.com/2017/06/07/installing-wireguard-the-modern-vpn/).
 
-This role was tested with Ubuntu 18.04 (Bionic Beaver), Debian 9 (Stretch) and Archlinux. It might also work with Ubuntu 16.04 (Xenial Xerus) but haven't tested it. If someone tested it let me please know if it works ;-)
+This role was tested with Ubuntu 18.04 (Bionic Beaver), Debian 9 (Stretch), Archlinux, Fedora 31 and CentOS. It might also work with Ubuntu 16.04 (Xenial Xerus) and Debian 10 (Buster) but haven't tested it. If someone tested it let me please know if it works ;-)
 
 Versions
 --------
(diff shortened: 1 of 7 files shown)


SiM22:
This would be a very very useful addition, needs to be merged!

zinefer:
I am using this and it seems to work well for me.

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
   tags:
     - wg-config
 
+- name: Create private key for unmanaged hosts
+  shell: "wg genkey | tee {{ wireguard_remote_directory }}/{{ item.host }}-privatekey"
+  args:
+    creates: "{{ wireguard_remote_directory }}/{{ item.host }}-privatekey"
+  register: uh_privkey
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
```
githubixx, line 115:
```suggestion
  with_items: "{{ wireguard_unmanaged_hosts }}"
```
If you put everything in a `block` as suggested above then there is no need for a default value as the block is only executed if `wireguard_unmanaged_hosts` is defined.

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+
+- name: Validate permissions of unmanaged hosts' private keys
+  file:
+    path: "{{ wireguard_remote_directory }}/{{ item.host }}-privatekey"
+    mode: '0400'
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
```
githubixx, line 121:
```suggestion
  with_items: "{{ wireguard_unmanaged_hosts }}"
```

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+
+- name: Recover existing private key for unmanaged hosts
+  shell: "cat {{ wireguard_remote_directory }}/{{ item.host }}-privatekey"
+  register: uh_privkey
+  changed_when: false
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
```
githubixx, line 127:
```suggestion
  with_items: "{{ wireguard_unmanaged_hosts }}"
```

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+
+- name: Derive WireGuard public key for unmanaged hosts
+  shell: "cat {{ wireguard_remote_directory }}/{{ item.host }}-privatekey | wg pubkey | tee {{ wireguard_remote_directory }}/{{ item.host }}-pubkey"
+  args:
+    creates: "{{ wireguard_remote_directory }}/{{ item.host }}-pubkey"
+  register: uh_pubkey
```
githubixx, line 133:
```suggestion
  register: uh_public_key
```
Not need to shorten variable name.

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
```
githubixx, line 134:
```suggestion
  with_items: "{{ wireguard_unmanaged_hosts }}"
```

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+
+- name: Recover existing public key for unmanaged hosts
+  shell: "cat {{ wireguard_remote_directory }}/{{ item.host }}-pubkey"
+  register: uh_pubkey
```
githubixx, line 138:
```suggestion
  register: uh_public_key
```

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+  changed_when: false
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
```
githubixx, line 140:
```suggestion
  with_items: "{{ wireguard_unmanaged_hosts }}"
```

tasks/main.yml
```diff
@@ -118,6 +151,15 @@
...
     - reconfigure wireguard
 
+- name: Generate WireGuard configuration file for unmanaged systems
+  template:
+    src: wg-unmanaged.conf.j2
+    dest: "{{ wireguard_remote_directory }}/{{ item.item.host }}.conf"
+    owner: root
+    group: root
+    mode: 0600
+  with_items: "{{ uh_privkey.results }}"
```
githubixx, line 161:
```suggestion
  with_items: "{{ uh_private_key.results }}"
```

README.md
```diff
@@ -257,6 +257,19 @@ Endpoint = server.at.home.p.domain.tld:51820
 
 The other WireGuard config files (`wg0.conf` by default) looks similar but of course `[Interface]` includes the config of that specific host and the `[Peer]` entries lists the config of the other hosts.
 
+Following variable allows to configure unmanaged hosts which are not configured by Ansible, typically smartphones and tablets.
+In this case, private key and client configuration are generated on wireguard hub.
+They can be shared through qrcode.
```
githubixx, line 262:
```suggestion
The following variable allows to configure unmanaged hosts which are not configured by Ansible, typically smartphones and tablets. In this case, private key and client configuration are generated on Wireguard hub. They can be shared through qrcode (also see https://fukuchi.org/works/qrencode/).

Example: `qrencode --type=ANSIUTF8 < /etc/wireguard/{{ item.item.host }}.conf`

```

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
+
+- name: Recover existing private key for unmanaged hosts
+  shell: "cat {{ wireguard_remote_directory }}/{{ item.host }}-privatekey"
+  register: uh_privkey
+  changed_when: false
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
+
+- name: Derive WireGuard public key for unmanaged hosts
+  shell: "cat {{ wireguard_remote_directory }}/{{ item.host }}-privatekey | wg pubkey | tee {{ wireguard_remote_directory }}/{{ item.host }}-pubkey"
```
githubixx, line 130:
```suggestion
  shell: |
    set -o errexit
    set -o pipefail
    cat {{ wireguard_remote_directory }}/{{ item.host }}-privatekey | wg pubkey | tee {{ wireguard_remote_directory }}/{{ item.host }}-pubkey
    exit 0
```

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+  args:
+    creates: "{{ wireguard_remote_directory }}/{{ item.host }}-pubkey"
```
githubixx, line 132:
```suggestion
    creates: "{{ wireguard_remote_directory }}/{{ item.host }}-pubkey"
    executable: "/bin/bash"
```

templates/wg.conf.j2
```diff
@@ -68,3 +68,11 @@ SaveConfig = true
     {% endif %}
   {% endif %}
 {% endfor %}
+{# unmanaged hosts #}
+{% for hostdata in uh_pubkey.results %}
```
githubixx, line 72:
```suggestion
{% for hostdata in uh_public_key.results %}
```

templates/wg.conf.j2
```diff
@@ -68,3 +68,11 @@ SaveConfig = true
     {% endif %}
   {% endif %}
 {% endfor %}
+{# unmanaged hosts #}
```
githubixx, line 71:
```suggestion
{% if wireguard_unmanaged_hosts is defined %}
{# unmanaged hosts #}
```

templates/wg.conf.j2
```diff
@@ -68,3 +68,11 @@ SaveConfig = true
...
+{% for hostdata in uh_pubkey.results %}
+
+    [Peer]
+    # {{ hostdata.item.host }}
+    PublicKey = {{ hostdata.stdout }}
+    AllowedIPs = {{ hostdata.item.allowed_ips}}
```
githubixx, line 77:
```suggestion
    AllowedIPs = {{ hostdata.item.allowed_ips}}
{% endif %}
```

tasks/main.yml
```diff
@@ -106,6 +107,38 @@
...
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
+
+- name: Validate permissions of unmanaged hosts' private keys
+  file:
+    path: "{{ wireguard_remote_directory }}/{{ item.host }}-privatekey"
+    mode: '0400'
+  with_items: "{{ wireguard_unmanaged_hosts | default([]) }}"
+
+- name: Recover existing private key for unmanaged hosts
+  shell: "cat {{ wireguard_remote_directory }}/{{ item.host }}-privatekey"
```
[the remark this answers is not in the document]
juju4: my only concern here is uh_pubkey can be fill either by key generation, either by key recovery.
if using slurp, output format will change and more processing of data after.
advice?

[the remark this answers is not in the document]
githubixx: TBH I'm not sure if I understand your question. This task just assigns a variable value. The needed base64 encoding/decoding is not worth talking about IMHO. If you would run this task 10x / sec. it would be different of course. Forking a shell also costs resources maybe even more then the Jinja "b64decode" filter.

tasks/main.yml
```diff
@@ -97,6 +97,45 @@
   tags:
     - wg-config
 
+- block:
+    - name: Create private key for unmanaged hosts
+      shell: |
+        set -o errexit
+        set -o pipefail
```
zinefer, line 104: I suddenly started failing with a `Illegal option -o pipefail`. Adding an `executable: /bin/bash` fixed the issue for me.

juju4: added

tasks/main.yml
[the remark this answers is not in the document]
juju4: uh_private_key is output of two tasks:
* Create private key for unmanaged host: change if priv key file does not exist
* Recover existing private key for unmanaged hosts. change_when: false
for which format is the same.

if moving to slurp, we get key only from latter. is that what is wanted?

Legogris:
My 5c as user:
Would be great if the auto-key-generation was optional (ie public key definable, together with endpoint). This would allow to define additional "upstreams" or disconnect plays between separated networks that should still be meshed.

That being said, for the intended feature-set, this looks mergable to me!

githubixx:
Sorry for the late answer. I was really short on free time to work on that. I'm sorry but I won't merge that PR (at least not as it is currently implemented). But I definitely see the need for this feature just to make that sure. After I saw it the first time I already suspected that it won't work as it should and I was unhappy that managed and un-managed hosts are handled quite differently. Now after testing it a while ago and thinking about it I decided not to merge it as I need to maintain it and I don't like the implementation. I should have done the testing before the code review. I'm really sorry for the unneeded work!

Currently it would create a file with the private key for every un-manged host on every other host. So you basically distribute secrets to all managed hosts. That's not how it should be. E.g. in my case I manage around twelve hosts with this role. If I now add my mobile e.g. as un-manged host then this PR would create a file for my mobile with a different private key on every host if I remember correctly. So all that managed hosts now have a secret stored that don't really care about at all because they don't need it and they should just not have it.

So I wonder where should the private key/configuration of the un-manged hosts be stored? IMHO the only location is the host at which `ansible-playbook` runs. That could either be a laptop/PC or a Jenkins server or something like that (maybe the configuration files are even located on a network share that such a host has mounted). The perfect solution would be of course to directly store the generated private keys in a vault like Hashicorp's Vault or Ansible Vault but that's a different story :wink: So for now I think the the only location is a local directory but definitely not the managed hosts. Also the private keys for un-managed devices should be generated on the host which executes `ansible-playbook`.

I've implemented a very, very, very basic skeleton how I would start implementing it: https://github.com/githubixx/ansible-role-wireguard/compare/unmanaged-hosts?expand=1
I've currently no idea how the final implementation would look like. But I think this would make it possible to handle a un-managed host more like a managed one so to say. So it would be possible to use at least a subset of the variables that can already be defined for managed hosts like `wireguard_dns`, `wireguard_allowed_ips`, and so on.

What's currently missing in the branch mentioned above is of course the most important part: As now all private keys or the configuration files (to be more precise) are located locally they need to be read and parsed from that host (`delegate_to: localhost`) like it is done with the managed hosts (https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/main.yml#L64-L90) to get the private key and derive the public key of the un-managed device. The further handling of these un-managed devices configuration would now require special handling of course. To store all variables for a un-managed device it might be necessary to create dynamic variable names to loop over later in the template by iterating through `wireguard_unmanaged_hosts_list` or use an appropriate data structure like an array or list which stores dictionaries e.g. Needs to be figured out while implementing it.

So that's just ideas that came into my mind. If you still have the power and willingness to work on it please let me know. Otherwise I'll close this chapter for now.

joneskoo:
@githubixx How about simpler; just add option to set variable for all hosts (or specific one) to dump plain WireGuard config snippet in? That way ansible-role-wireguard would only update the config on ansible-managed hosts but how that peer config is generated is completely external (i.e. no need to care how private keys are handled). If I'm not missing something, that should be just adding a template variable into the j2 template, and injecting the peer config. Could be a local file if it's awkward to inline config in Ansible YAML.

githubixx:
I think this is a dead end for now so I'll close the PR. https://github.com/githubixx/ansible-role-wireguard/pull/63 seems to be a better way to have basic support for unmanged hosts.

PR #46, opened 2020-03-24
Forcing stop and start to clear routes

cristichiru:
Rationale:
- At least on Ubuntu 18.04, if running this role multiple times, additional routes are not being set
- A full stop and start works well, compared to just a service restart


cristichiru:
Hmm, good to know, I will check my wg version. Thanks!

PR #48, opened 2020-04-27
Use Arch Linux specific package list

juergenhoetzel:
Arch Linux ships a Linux kernel > 5.6 and doesn't require DKMS.

Move the package list variable to (distribution-specific) var files.

For the Arch Linux LTS kernel (5.4) a binary wireguard-lts package is
provided in [core].

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -4,14 +4,20 @@
 
 - include_tasks: "setup-{{ ansible_distribution|lower }}.yml"
 
+- name: Load packages variable file based on the OS type, or a default if not found
+  include_vars: "{{ lookup('first_found', params) }}"
+  vars:
+    params:
+      files:
+        - "packages-{{ ansible_distribution | lower }}.yml"
+        - "packages.yml"
+      paths:
+        - "vars"
+
 - name: Install WireGuard
   package:
     name: "{{ packages }}"
     state: present
-  vars:
-    packages:
-    - wireguard-dkms
-    - wireguard-tools
   tags:
     - wg-install
 
--- a/tasks/setup-archlinux.yml
+++ b/tasks/setup-archlinux.yml
@@ -1,11 +1,23 @@
 ---
-- name: Install required packages
+- name: Install wireguard-lts package
   pacman:
-    name: "{{ packages }}"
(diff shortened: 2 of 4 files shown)


tasks/main.yml
```diff
@@ -4,14 +4,20 @@
 
 - include_tasks: "setup-{{ ansible_distribution|lower }}.yml"
 
+- name: Load packages variable file based on the OS type, or a default if not found
+  include_vars: "{{ lookup('first_found', params) }}"
+  vars:
+    params:
+      files:
+        - "packages-{{ ansible_distribution }}.yml"
```
githubixx, line 12:
```suggestion
        - "packages-{{ ansible_distribution|lower }}.yml"
```
This way `vars/packages-Archlinux.yml` can be renamed to `vars/packages-archlinux.yml`.

tasks/setup-archlinux.yml
```diff
@@ -6,6 +6,7 @@
   become: yes
   vars:
     packages:
-      - linux-headers
+      - wireguard-lts
   tags:
     - wg-install
+  when: ansible_kernel is match(".*-lts$")
```
githubixx, line 12: Not sure if this is the right approach here. Currently the LTS kernel is 5.4 but sooner or later it will become a greater  version. So just checking for "-lts" might not be good enough. Maybe additionally check if the kernel >= 5.6 like described here: https://docs.ansible.com/ansible/latest/user_guide/playbooks_tests.html#version-comparison 

Also people using this role for a while most probably already have installed `wireguard-dkms` and need an upgrade path. So a task is needed to remove it as it will collide with `wireguard-lts`.

Additionally `wireguard-dkms` is still needed for people with custom kernel or people who use `linux-mainline` from AUR e.g. There is also a package `extra/linux-zen` which is currently kernel `5.6.7`.

So to handle this maybe a `block` statement makes sense like used here: https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/main.yml#L59-L71

juergenhoetzel: Thanks a lot for your helpful feedback: I have addressed your feedback in the amended commit!

PR #49, opened 2020-04-27
Use wireguard from buster-backports on debian if needed

ties:
I found out that my changes to add debian support no longer worked.

  * Wireguard key was missing from keyserver and no longer needed to be added for buster-backports.
  * `buster-backports` contains wireguard, using this for <= buster
  * no longer pin packages from `unstable`. No need to pin backports, it has a lower priority than main repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Changelog
 ---------
 
-*6.0.4*
+**6.0.4**
 
 - Use the buster-backports repository on Debian Buster (or older), use package
   standard repositories on sid/bullseye.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+*6.0.4*
+
+- Use the buster-backports repository on Debian Buster (or older), use package
+  standard repositories on sid/bullseye.
+
 **6.0.3**
 
 - If `wg syncconf` command is not available do stop/start service instead of restart (contribution by @cristichiru)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ Changelog
 - Use the buster-backports repository on Debian Buster (or older), use package
   standard repositories on sid/bullseye.
 
+  The role no longer adds the unstable _repo_ nor the _apt preference_ for that repo.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+*6.0.4*
```
githubixx, line 4:
```suggestion
**6.0.4**
```

tasks/setup-debian.yml
```diff
@@ -1,32 +1,15 @@
 ---
-- name: Setup WireGuard preference
-  copy:
-    src: debian/etc/apt/preferences.d/limit-unstable
-    dest: /etc/apt/preferences.d/limit-unstable
```
githubixx, line 5: I guess we can assume that this file already exists if a user used that role already. So it should be mentioned at least in the changelog to delete this file manually or use a task to delete it.

tasks/setup-debian.yml
```diff
@@ -1,32 +1,15 @@
...
-    keyserver: "keyserver.ubuntu.com"
-    id: "8B48AD6246925553"
-    state: present
-  tags:
-    - wg-install
-
-- name: Add WireGuard repository
+- name: Add WireGuard repository on buster or earlier
   apt_repository:
-    repo: "deb http://deb.debian.org/debian/ unstable main"
```
githubixx, line 27: I guess this task added a file in `/etc/apt/sources.list.d/`. So it will be still there after this change. Maybe this should also be mentioned in the changelog.

tasks/setup-debian.yml
ties: It is a scary one. If you remove the pin but not the repo, you will start updating from unstable (a bad idea).

If you remove the repo + the pin that will work. I'll add this to the changelog.

CHANGELOG.md
ties: Funny, my vim was hiding the double stars but not on the line itself. So they looked similar to me…

tasks/setup-debian.yml
ties: Good that you caught these things in the review. I think I clarified it sufficiently, but let me know if it needs further clarification.

Issue #53, opened 2020-05-25
Linux header installation fails on Raspberry pi


penguineer:
Installation fails on Raspberry Pi with Raspbian : setup-debian.yml tries to install `"linux-headers-{{ dpkg_arch.stdout }}"`, however, in Raspbian the package is called linux-headers (without arch).

penguineer:
Possible solutions:
* Make Raspbian provide a package linux-headers-armhf
* Create a role setup-raspbian that takes the Raspbian anomaly into account
* Change the Debian-Role to deal with missing arch headers and try `linux-headers` instead

penguineer:
Apparently the Debian keys also need to be added, which is an argument towards adding a Raspbian-specific role

githubixx:
As I don't use a Raspberry it's unlikely that I'll do any changes in this regard. But pull requests are welcome. I guess too that a Raspbian specific role would make sense.

penguineer:
Thanks for your reply!

Apparently there are two issues at hand:
* Raspbian is identified as Debian (with the assumption, that `ansible_distribution` identifies the target system)
* but it does not provide `"linux-headers-{{ dpkg_arch.stdout }}"`

I tried to go with the following in the setup-debian role:
```yaml
- name: (Debian) Install kernel headers to compile Wireguard with DKMS
  block:
    - apt:
        name:
        - "linux-headers-{{ dpkg_arch.stdout }}"
        state: present
  rescue:
    - apt:
        name:
        - "linux-headers"
        state: present
```

However, this will still report a failure, as expected from the block/rescue construct (even though the result is installed kernel headers on both systems).

I will try to fix the issues on  the Raspbian side and/or find a less intrusive way to adapt the setup-debian role.

githubixx:
As the PR was merged I guess we can close this issue.

PR #54, opened 2020-06-01
Add a role for Raspbian

penguineer:
Raspbian is a Debian flavor, but has some deviations. The debian-raspbian role sets up WireGuard on a Raspberry Pi with a recent raspbian:
* Setup Debian unstable repository
* Update the kernel and reboot, if necessary
* Install wireguard-dkms and  -tools

The switch in the `debian-setup` role is necessary because Debian and Raspbian are identified as "Debian".

This is a fix for #53.


penguineer:
Thank you very much!

PR #55, opened 2020-06-03
Fix on debian openstack images

pallinger:
On openstack Debian images, the kernel is different, so we need to install different kernel headers, too.

### DIFF
--- a/tasks/setup-debian.yml
+++ b/tasks/setup-debian.yml
@@ -18,10 +18,16 @@
   register: dpkg_arch
   changed_when: False
 
+- name: (Debian) Get kernel version
+  command: "uname -r"
+  register: kernel_version
+  changed_when: False
+
 - name: (Debian) Install kernel headers to compile Wireguard with DKMS
   apt:
     name:
       - "linux-headers-{{ dpkg_arch.stdout }}"
+      - "linux-headers-{{ kernel_version.stdout }}"
     state: present
 
 - name: (Debian) Install wireguard packages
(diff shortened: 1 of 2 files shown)


tasks/setup-debian.yml
```diff
@@ -18,10 +18,16 @@
...
+- name: (Debian) Get kernel version
+  command: "uname -r"
+  register: kernel_version
+  changed_when: False
+
 - name: (Debian) Install kernel headers to compile Wireguard with DKMS
   apt:
     name:
       - "linux-headers-{{ dpkg_arch.stdout }}"
+      - "linux-headers-{{ kernel_version.stdout }}"
```
githubixx, line 30: I just merged https://github.com/githubixx/ansible-role-wireguard/pull/54. So you need to solve some conflicts. But I'm not sure for what that is needed. `linux-headers-amd64` is a meta-package e.g. This will install `linux-headers-4.19.0-9-amd64` and `linux-headers-4.19.0-9-common` at the moment. So for a "vanilla/normal" Debian this patch isn't needed. If this is needed for Openstack it would make sense to have a separate task with a `when:` clause. I would suggest to run `ansible -m setup <hostname>` for an example host with Openstack installed and check the output if it contains something Openstack specific which can be checked in a `when:` clause. But still I'm not sure if this patch is really needed.

pallinger: Ok, I merged your master branch and made the different Linux header version conditional. Please take a look now!

pallinger: Sorry, I made a syntax error :(, please wait!

pallinger: Ok, the syntax error is gone, ready for review!

PR #58, opened 2020-07-16
Add wireguard_preshared_key variable

Legogris:
It could be desired to use a preshared key for additional security. This accommodates for that by an additional role variable `wireguard_preshared_key`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Changelog
 ---------
 
+**6.4.0**
+
+- Support specifying optional `wireguard_preshared_key` (contribution by @legogris)
+
 **6.3.1**
 
 - Support Openstack Debian images (contribution by @pallinger)
--- a/README.md
+++ b/README.md
@@ -38,6 +38,9 @@ wireguard_port: "51820"
 
 # The default interface name that wireguard should use if not specified otherwise.
 wireguard_interface: "wg0"
+
+# Preshared key to use for all peers (optional)
+wireguard_preshared_key: "abc123"
 ```
 
 The following variable is mandatory and needs to be configured for every host in `host_vars/`:
--- a/templates/wg.conf.j2
+++ b/templates/wg.conf.j2
@@ -50,6 +50,9 @@ SaveConfig = true
     {% else %}
     AllowedIPs = {{hostvars[host].wireguard_ip}}/32
     {% endif %}
+    {% if wireguard_preshared_key is defined %}
+    PresharedKey = {{wireguard_preshared_key}}
+    {% endif %}
     {% if hostvars[host].wireguard_persistent_keepalive is defined %}
     PersistentKeepalive = {{hostvars[host].wireguard_persistent_keepalive}}
     {% endif %}


githubixx:
Thanks for the contribution! But the problem here is that the implementation is okay for a client/server setup (one client / one server) but it's not sufficient if you have more peers. In general it works but as [Archlinux Wiki](https://wiki.archlinux.org/index.php/WireGuard#Key_generation) states you should have one pre-shared key for each peer pair:

```
One can also generate a pre-shared key to add an additional layer of symmetric-key cryptography to be mixed into the already existing public-key cryptography, for post-quantum resistance. A pre-shared key should be generated for each peer pair and should not be reused. For example, three interconnected peers, A, B, and, C will need three separate pre-shared keys, one for each peer pair.

Generate a pre-shared key for each peer pair using the following command:

$ wg genpsk > peer_A-peer_B.psk
$ wg genpsk > peer_A-peer_C.psk
$ wg genpsk > peer_B-peer_C.psk
```
Of course if you use only one PSK for all peers that increases security in general but I guess people would expect that this is implemented properly. And in this case `wireguard_preshared_key` wouldn't be a "global" variable anymore but a "per peer pair" variable so to say.

I'm not sure if this should be merged if it'll stays as is ATM. This is definitely something that must be documented in the `wireguard_preshared_key` comment. And maybe even the variable should be renamed to `wireguard_preshared_key_for_all_peers` to a) make it possible to implement the peer pair PSK thingy mentioned above later and b) to make it clear that this PSK is shared by all peers.

Also the role currently doesn't store any secrets (public/private keys) besides in the WireGuard configuration files on the WireGuard peers. The private keys are all rendered on the WireGuard peers and stored there. The public keys are derived from the private keys if needed. So instead of a variable with the PSK value it would make more sense to delegate the PSK generation to one host (e.g. the first one of the current play), store the PSK in a variable during the run and then distribute it to all hosts in case of a "global" PSK for all WireGuard peers. 

In case of a PSK per peer pair (or even worse in case you've a mesh where every peer can communicate with all the other peers) PSK generation needs to be delegated to the WireGuard peers anyways and also the PSK should be stored there and fetched during playbook run so that it can be used in templates e.g. Storing it locally doesn't makes that much sense. I did this at the very beginning I developed this role but that wasn't really considered best practice :wink: 

So I think it's a little bit more complicated then just adding a variable IMHO esp. if more then two peers should be able communicate.

Legogris:
@githubixx Good point on this encouraging bad practices by making it a global variable, it certainly gives the wrong idea to users. Do you think it could make sense to make it a host-level argument?

I realize this role is built around the scenario where all peers are managed by the role, or at least by the person operating it - especially re key generation.

However, I think a commonly desired use-case is one where this is either not the case or not desired (consider for example dynamically provisioned containers connecting from Kubernetes/Nomad/Swarm/Mesos, having their keys managed in a secret storage. Or peers living on edges routers not managable by ansible. Or when connecting with foreign networks and psk/pubkey fora peer is outside our control. Or configurations for end-users. etc). TBH I think it'd make more sense to have keys provided as role variables (when managed properly through e.g. Ansible Vault) than generated on run-time, but I see both the convenience factor and the benefit of never having the pk leave the peer here.

Either way, personally my main use-case for this would be together with https://github.com/githubixx/ansible-role-wireguard/pull/41, and actually setting it to a unique value for each host in inventory, but I split it out to make the feature/change independent.

That being said, totally see why this probably shouldn't be merged as-is given the above.

joneskoo:
@Legogris Foreign peers unmanaged by ansible-role is being implemented in #63 .

My 2 cents: I think adding psk for this role only makes sense if it's set up between each pair. Probably wouldn't need configuration at all, just silently implemented when peer-to-peer PSK generation is implemented. I'd archive this PR without merging.

githubixx:
As joneskoo suggested I'll close this for now. It really only makes sense to set it up between each host pair.

Issue #50, opened 2020-05-01
Switch to ELRepo for RHEL/Centos


elcomtik:
I'm using this role to install wireguard on Centos7/8 systems. In recent update to Centos 7.8 wireguard caused problems(kernel panic at boot).  It updated kernel, however wireguard kernel module wasn't updated.

I troubleshooted issue and found that this role installed dkms-wireguard package from jdoss repository, which seems to be not maintained properly. I found that official wireguard webpage recommends to use ELRepo and kmod-wireguard package https://www.wireguard.com/install/. I manually switched to this repository and it worked perfectly. Extra bonus for getting rid of dependencies for building dkms.

We should also add some task to properly uninstall old repository and dkms package.

I would make PR if you would like.

githubixx:
That would be great if you could submit a PR. The initial PR for CentOS support was created by a contributor as I don't use that distribution. The PR was LGTM at that time.

PR #59, opened 2020-08-05
Switched to ELRepo for Centos

elcomtik:
This PR solves #50 

- added switch to differentiate setup of Centos7/8
- replaced old repository by officialy recomended
- added step to remove old dkms wireguard package
- switched to install KMOD wireguard package


githubixx:
Awsome, thanks! :smiley: I'll have a look as soon as possible. Can you please also update CHANGELOG accordingly? Maybe also add some information about breaking changes that a user of the role need to consider before upgrading. Thanks!

elcomtik:
I' will update CHANGELOG soon.

I have a dilemma about semantic versioning. This change may break installation for systems with custom kernels(no idea if there is someone who needs it). The role previously supported custom kernel implicitly because it was using DKMS package. I consider this feature out of the scope of this role until it's needed. What version number would you propose to use?

githubixx:
I would use a new major version. So `7.0.0` that would be. I'm not so strict about semantic versioning. But if it could potentially break something then a new major version definitely makes sense IMHO. That should users at least remind to read the changelog before upgrading :wink: And if someone figures out that she/he is not affected then it won't hurt anyone. But for affected users it's important to know what could happen.

For Archlinux we had basically the same situation. There you've linux, linux-lts, linux-zen, ... WireGuard module was officially included in kernel 5.6. So in that case no DKMS is used for everything else DKMS is still used. That's also similar to Ubuntu which uses kernel 5.4 for Ubuntu 20.04. But in that case they backported WireGuard to kernel 5.4...

elcomtik:
Ok, job done. You may review it.

CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 Changelog
 ---------
 
+**7.0.0**
+
+- Switched to install from ELRepo KMOD package for Centos. This change may break installation for systems with custom kernels. The role previously supported custom kernel implicitly because it was using DKMS package.
```
githubixx, line 6:
```suggestion
- Switched to install from ELRepo KMOD package for CentOS (see https://www.wireguard.com/install/). This change may break installation for systems with custom kernels. The role previously supported custom kernel implicitly because it was using DKMS package (contribution by @elcomtik)
```

elcomtik:
> Your PR removes the DKMS only but doesn't touch the old repository. I don't know if it's a good idea to remove the old one automatically or if we should leave it to the user. In the later case we should at least mention it in the CHANGELOG and if it should be done before or after the updated role was rolled out. That would be nice to clarify.

I would let the user to decide whether to remove the old repository, because it can be used by other software. Note in CHANGELOG should be sufficient.

elcomtik:
I think it's ready to ship

PR #61, opened 2020-08-17
Add basic macOS support

rubendibattista:
I added basic support for `macOS`

- Add the `ansible_os_is_linux` internal variable to discern between a Linux system and a macOS one
- Conditional choice for `wireguard_remote_directory` default value. It defaults to `/etc/wireguard` on Linux and `/opt/local/etc/wireguard` (i.e. Macports default path) on macOS
- Add configurable `owner` and `group` for the configuration file `wg0.conf`. The group defaults on `wheel` on `macOS`
- No handlers for macOS 
- No kernel module activation on macOS
- Add instructions into the `README.md` on how to run the VPN on macOS manually

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -4,13 +4,22 @@
 #######################################
 
 # Directory to store WireGuard configuration on the remote hosts
-wireguard_remote_directory: "/etc/wireguard"
+wireguard_remote_directory: "{{ '/etc/wireguard' if not ansible_os_family == 'Darwin' else '/opt/local/etc/wireguard' }}"
 
 # The default port WireGuard will listen if not specified otherwise.
 wireguard_port: "51820"
 
 # The default interface name that wireguard should use if not specified otherwise.
 wireguard_interface: "wg0"
+
+# The default owner of the wg.conf file
+wireguard_conf_owner: root
+
+# The default group of the wg.conf file
+wireguard_conf_group: "{{ 'root' if not ansible_os_family == 'Darwin' else 'wheel' }}"
+
+# The default mode of the wg.conf file
+wireguard_conf_mode: 0600
 
 
 #######################################
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -6,7 +6,9 @@
   loop:
   - stopped
   - started
-  when: not wg_syncconf
+  when: >
+    not wg_syncconf and
+    not ansible_os_family == 'Darwin'
   listen: "reconfigure wireguard"
 
 - name: syncconf wireguard
@@ -19,5 +21,7 @@
       exit 0
   args:
     executable: "/bin/bash"
-  when: wg_syncconf
+  when: >
+    wg_syncconf and
+    not ansible_os_family == 'Darwin'
   listen: "reconfigure wireguard"
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -10,17 +10,19 @@
     - "setup-{{ ansible_distribution|lower }}.yml"
     - "setup-{{ ansible_os_family|lower }}.yml"
 
-- name: Enable WireGuard kernel module
-  modprobe:
-    name: wireguard
-    state: present
-  register: wireguard_module_enabled
-  until:  wireguard_module_enabled is succeeded
-  retries: 10
-  delay: 10
-  failed_when: wireguard_module_enabled is failure
-  tags:
-    - wg-install
+- block:
+    - name: Enable WireGuard kernel module
+      modprobe:
+        name: wireguard
+        state: present
+      register: wireguard_module_enabled
+      until:  wireguard_module_enabled is succeeded
+      retries: 10
+      delay: 10
+      failed_when: wireguard_module_enabled is failure
+      tags:
+        - wg-install
+  when: not ansible_os_family == 'Darwin'
 
 - name: Set WireGuard IP (without mask)
   set_fact:
@@ -107,9 +109,9 @@
   template:
     src: wg.conf.j2
     dest: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
-    owner: root
-    group: root
-    mode: 0600
+    owner: "{{ wireguard_conf_owner }}"
+    group: "{{ wireguard_conf_group }}"
+    mode: "{{ wireguard_conf_mode }}"
   tags:
     - wg-config
   notify:
@@ -135,3 +137,4 @@
     name: "wg-quick@{{ wireguard_interface }}"
     state: started
     enabled: yes
+  when: not ansible_os_family == 'Darwin'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 Changelog
 ---------
+**7.2.0**
+
+- Basic MacOS X support (contribution by @rubendibattista)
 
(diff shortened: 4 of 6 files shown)


vars/main.yml
```diff
@@ -0,0 +1,2 @@
+# This variable is true if the host OS is linux
+ansible_os_is_linux: "{{ ansible_os_family in ['Debian', 'ArchLinux', 'RedHat'] }}"
```
githubixx, line 2:
```suggestion
ansible_os_is_linux: "{{ ansible_os_family in ['Debian', 'Archlinux', 'RedHat'] }}"
```
According to https://docs.ansible.com/ansible/latest/user_guide/playbooks_conditionals.html#ansible-facts-os-family `Archlinux` is the correct value (and also according to my PC :wink: ).

tasks/linux.yml
```diff
@@ -0,0 +1,11 @@
+- name: Enable WireGuard kernel module
```
githubixx, line 1: You don't need an extra file. Leave it were it's currently and put it in a `block:` with a condition `when: ansible_os_is_linux`. Here is an example how this looks like: https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/main.yml#L55-L68

githubixx (changes requested):
Thanks for you contribution! Comments inline. Please also add an entry to CHANGELOG.md e.g.:
```
**7.1.0**

- Basic MacOS X support (contribution by @rubendibattista)
```

githubixx:
@rubendibattista Are you still working on this? Otherwise I would just close the PR to clean up the PRs in the queue.

rubendibattista:
Ehi, I forgot this.. 😔. I'll apply your suggestions asap.

githubixx:
@rubendibattista Thanks for your feedback! Maybe keep an eye on https://github.com/githubixx/ansible-role-wireguard/pull/63. I guess I'll merge that one first. I think there shouldn't be too much merge conflicts but just that you're aware that there is something going on.

rubendibattista:
Ehi @githubixx, I should have applied the requested changes. 👍 

If you merge before #63, I'll rebase on it and if there're conflicts, try to fix them.

githubixx:
@rubendibattista #63 is merged now. So please rebase. Looks like there are no conflicts :smiley:

rubendibattista:
It's done!

defaults/main.yml
```diff
@@ -4,14 +4,23 @@
 #######################################
 
 # Directory to store WireGuard configuration on the remote hosts
-wireguard_remote_directory: "/etc/wireguard"
+wireguard_remote_directory: "{{ '/etc/wireguard' if ansible_os_is_linux else '/opt/local/etc/wireguard' }}"
```
githubixx, line 7:
```suggestion
wireguard_remote_directory: "{{ '/etc/wireguard' if not ansible_os_family == 'Darwin' else '/opt/local/etc/wireguard' }}"
```

defaults/main.yml
```diff
@@ -4,14 +4,23 @@
...
 wireguard_port: "51820"
 
 # The default interface name that wireguard should use if not specified otherwise.
 wireguard_interface: "wg0"
 
+# The default owner of the wg.conf file
+wireguard_conf_owner: root
+
+# The default group of the wg.conf file
+wireguard_conf_group: "{{ 'root' if ansible_os_is_linux else 'wheel' }}"
```
githubixx, line 19:
```suggestion
wireguard_conf_group: "{{ 'root' if not ansible_os_family == 'Darwin' else 'wheel' }}"
```

handlers/main.yml
```diff
@@ -6,7 +6,7 @@
   loop:
   - stopped
   - started
-  when: not wg_syncconf
+  when: not wg_syncconf and ansible_os_is_linux
```
githubixx, line 9:
```suggestion
  when: >
    not wg_syncconf and
    not ansible_os_family == 'Darwin'
```

handlers/main.yml
```diff
@@ -19,5 +19,5 @@
       exit 0
   args:
     executable: "/bin/bash"
-  when: wg_syncconf
+  when: wg_syncconf and ansible_os_is_linux
```
githubixx, line 22:
```suggestion
  when: >
    wg_syncconf and
    not ansible_os_family == 'Darwin'
```

tasks/main.yml
```diff
@@ -10,17 +10,19 @@
...
+        name: wireguard
+        state: present
+      register: wireguard_module_enabled
+      until:  wireguard_module_enabled is succeeded
+      retries: 10
+      delay: 10
+      failed_when: wireguard_module_enabled is failure
+      tags:
+        - wg-install
+  when: ansible_os_is_linux
```
githubixx, line 25:
```suggestion
  when: not ansible_os_family == 'Darwin'
```

tasks/main.yml
```diff
@@ -135,3 +137,4 @@
     name: "wg-quick@{{ wireguard_interface }}"
     state: started
     enabled: yes
+  when: ansible_os_is_linux
```
githubixx, line 140:
```suggestion
  when: not ansible_os_family == 'Darwin'
```

vars/main.yml
```diff
@@ -0,0 +1,2 @@
+# This variable is true if the host OS is linux
```
githubixx, line 1: If the changes suggested above works out this file can be removed.

githubixx (changes requested):
@rubendibattista I added a few comments. I basically "inverted" the logic. While I don't support all Linux distributions possibility is high that this role runs with little to no modification on SuSE e.g. or basically on every Linux distribution that uses `systemd`. If I would apply your changes without modification those users are locked out or need even more changes to make it work. That's something I want to avoid.

In this case MacOS is the exception and not the rule so to say :wink: While not totally happy with all this `not`'s it still makes more sense at the end IMHO.

Please test the suggested changes and if it works out apply them if they're ok for you.

rubendibattista:
I should have applied the modifications and running the playbook on my system seems to work as expected.

githubixx (approved):
Thanks!

tasks/main.yml
```diff
@@ -10,17 +10,19 @@
...
+        name: wireguard
+        state: present
+      register: wireguard_module_enabled
+      until:  wireguard_module_enabled is succeeded
+      retries: 10
+      delay: 10
+      failed_when: wireguard_module_enabled is failure
+      tags:
+        - wg-install
+  when: not ansible_os_family == 'Darwin'
```
ypid, line 25: Why did you wrap this one task in a block?

rubendibattista: Because on macOS there's no kernel module to enable. You need to manually run the Go binary (or use the official App).

ypid: My question is just about the syntax :) The `when` condition can also directly be attached to this one `modprobe` task.

githubixx: AFAIR that was a suggestion by me. But you are right the `block` isn't needed. The code looked a little bit different at the beginning. That's an oversight of me.

rubendibattista: Oups, sorry. Should I open another PR to fix this?

ypid: I guess you can do that. That is probably cleaner.

PR #63, opened 2020-09-01
Add support for unmanaged WireGuard peers

### LINKED ISSUE #45: No support for roaming peers
There is no support for roaming peers because every node expects every-other peer to have a resolvable endpoint. Roaming peers (like a laptop) will never have resolvable endpoints so there must be a way to omit them from being present in other peer configs.

The issue comes from the config file template around this part.

```jinja2
{% for host in ansible_play_hosts %}
  {% if host != inventory_hostname %}
  
    [Peer]
```

Suggest adding a new variable `wireguard_roaming` to omit peers that are roaming.

```jinja2
{% for host in ansible_play_hosts %}
  {% if host != inventory_hostname %}
  
    [Peer]
    ...
    {% if hostvars[host].wireguard_roaming is undefined or not hostvars[host].wireguard_roaming %}
    ...
    {% endif %}
  {% endif %}
{% endfor %}
```

joneskoo:
In order to support peers that are not managed in Ansible, add support to add peers not managed by ansible to WireGuard config. This allows easily defining peers without endpoint that can talk to all nodes being deployed. Especially useful for workstations without an endpoint address.

Add variable wireguard_unmanaged_peers that appends the peers into config.

This closes #41, and closes #45.

Example of how to use this:

```yaml
vpn:
  hosts:
    server1:
      ...
    server2:
      ...
      wireguard_unmanaged_peers:
        client.example.com:
          public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
          allowed_ips: 10.0.0.3/32
  vars:
    wireguard_unmanaged_peers:
      client.example.com:
        public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
        allowed_ips: 10.0.0.3/32
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,7 +98,7 @@ wireguard_save_config: "true"
 wireguard_unmanaged_peers:
   client.example.com:
     public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
-    preshared_key: mfKkizs9J4eRFnyhaRvUKdFAxoGdDfcgMy4OhrbyjPE=
+    # preshared_key: ... e.g. from ansible-vault?
     allowed_ips: 10.0.0.3/32
     endpoint: client.example.com:51820
     persistent_keepalive: 0
--- a/README.md
+++ b/README.md
@@ -95,11 +95,13 @@ wireguard_postup:
 wireguard_postdown:
   - ...
 wireguard_save_config: "true"
-wireguard_extra_peer_config: |
-  # Configure external peers - not managed by ansible-role-wireguard.
-  [Peer]
-  PublicKey = 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
-  AllowedIPs = 10.0.0.3/32
+wireguard_unmanaged_peers:
+  client.example.com:
+    public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
+    preshared_key: mfKkizs9J4eRFnyhaRvUKdFAxoGdDfcgMy4OhrbyjPE=
+    allowed_ips: 10.0.0.3/32
+    endpoint: client.example.com:51820
+    persistent_keepalive: 0
 ```
 
 `wireguard_(preup|predown|postup|postdown)` are specified as lists. Here are two examples:
--- a/README.md
+++ b/README.md
@@ -95,6 +95,11 @@ wireguard_postup:
 wireguard_postdown:
   - ...
 wireguard_save_config: "true"
+wireguard_extra_peer_config: |
+  # Configure external peers - not managed by ansible-role-wireguard.
+  [Peer]
+  PublicKey = 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
+  AllowedIPs = 10.0.0.3/32
 ```
 
 `wireguard_(preup|predown|postup|postdown)` are specified as lists. Here are two examples:
--- a/templates/wg.conf.j2
+++ b/templates/wg.conf.j2
@@ -73,21 +73,21 @@ SaveConfig = true
     # Peers not managed by ansible from wireguard_unmanaged_peers
     {% for peer in wireguard_unmanaged_peers.keys() %}
     [Peer]
-    # {{ peer}}
+    # {{ peer }}
     {% if wireguard_unmanaged_peers[peer].public_key is defined %}
-    PublicKey = {{wireguard_unmanaged_peers[peer].public_key}}
+    PublicKey = {{ wireguard_unmanaged_peers[peer].public_key }}
     {% endif %}
     {% if wireguard_unmanaged_peers[peer].preshared_key is defined %}
-    PresharedKey = {{wireguard_unmanaged_peers[peer].preshared_key}}
+    PresharedKey = {{ wireguard_unmanaged_peers[peer].preshared_key }}
     {% endif %}
     {% if wireguard_unmanaged_peers[peer].allowed_ips is defined %}
-    AllowedIPs = {{wireguard_unmanaged_peers[peer].allowed_ips}}
+    AllowedIPs = {{ wireguard_unmanaged_peers[peer].allowed_ips }}
     {% endif %}
     {% if wireguard_unmanaged_peers[peer].endpoint is defined %}
-    Endpoint = {{wireguard_unmanaged_peers[peer].endpoint}}
+    Endpoint = {{ wireguard_unmanaged_peers[peer].endpoint }}
     {% endif %}
     {% if wireguard_unmanaged_peers[peer].persistent_keepalive is defined %}
-    PersistentKeepalive = {{wireguard_unmanaged_peers[peer].persistent_keepalive}}
+    PersistentKeepalive = {{ wireguard_unmanaged_peers[peer].persistent_keepalive }}
     {% endif %}
     {% endfor %}
 {% endif %}
--- a/templates/wg.conf.j2
+++ b/templates/wg.conf.j2
@@ -41,7 +41,7 @@ SaveConfig = true
 {% endif %}
 {% for host in ansible_play_hosts %}
   {% if host != inventory_hostname %}
-  
+
     [Peer]
     # {{ host }}
     PublicKey = {{hostvars[host].public_key}}
@@ -68,3 +68,7 @@ SaveConfig = true
     {% endif %}
   {% endif %}
 {% endfor %}
+{% if wireguard_extra_peer_config is defined %}
+    # Unmanaged peers from wireguard_extra_peer_config
+    {{wireguard_extra_peer_config|indent(4, False)}}
+{% endif %}
--- a/templates/wg.conf.j2
+++ b/templates/wg.conf.j2
@@ -74,9 +74,7 @@ SaveConfig = true
     {% for peer in wireguard_unmanaged_peers.keys() %}
     [Peer]
     # {{ peer }}
-    {% if wireguard_unmanaged_peers[peer].public_key is defined %}
     PublicKey = {{ wireguard_unmanaged_peers[peer].public_key }}
-    {% endif %}
     {% if wireguard_unmanaged_peers[peer].preshared_key is defined %}
     PresharedKey = {{ wireguard_unmanaged_peers[peer].preshared_key }}
     {% endif %}
--- a/templates/wg.conf.j2
+++ b/templates/wg.conf.j2
@@ -68,7 +68,26 @@ SaveConfig = true
     {% endif %}
   {% endif %}
 {% endfor %}
-{% if wireguard_extra_peer_config is defined %}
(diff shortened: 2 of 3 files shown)


joneskoo:
Based on quick test this is exactly what I need for the moment to take this role into use. Hopefully it solves a generic need. At least it's simple, and no-op by default.

joneskoo:
@githubixx  I saw you merged my other PR. I hope you don't mind a friendly ping as there was no comment if this approach would be welcome? No worries if you just didn't have time to look at it yet.

githubixx:
@joneskoo I didn't have time to test it, yet. And I need to think a little bit about it. I'll get back to you around end of next week I think.

j8r:
The idea is good, but it will be better to be a dictionary with `endpoint`, `public_key`, `allowed_ips`, `persistent_keepalive` as possible keys.

joneskoo:
> The idea is good, but it will be better to be a dictionary with `endpoint`, `public_key`, `allowed_ips`, `persistent_keepalive` as possible keys.

Why? What about preshared key and such?

joneskoo:
@j8r I updated the PR including description of original post based on your comment.

j8r:
Nice @joneskoo.
I was looking at https://docs.sweeting.me/s/wireguard#Peer, which did only mention pre-shared keys in another section.

What do you think of having a list of dict, with an other optional key `name`, which will be `# Name = {{name}}`?

Also, Endpoint and PublicKey ate mandatory, no? If so, no need to check if it they are defined – they must be.

joneskoo:
Hi @j8r. The dict of dict is symmetric with how hosts are defined in inventory. This role already writes the ansible host name of other hosts as comment for the peers generated so what I did matches that. Further, an explicit "name" makes you wonder if it's going to be written to the config so that's not a good name.

My version:
```yaml
vpn:
  hosts:
    server2:
      ...
      wireguard_unmanaged_peers:
        client.example.com:
          public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
          allowed_ips: 10.0.0.3/32
```

If I understand right you propose instead this:

```yaml
vpn:
  hosts:
    server2:
      ...
      wireguard_unmanaged_peers:
        - comment: client.example.com
          public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
          allowed_ips: 10.0.0.3/32
```

The config we generate looks like this:

```ini
[Interface]
# server2
Address = ...
PrivateKey = ...
ListenPort = 51820

    [Peer]
    # server1
    PublicKey = ...
    AllowedIPs = ...
    Endpoint = ...

    # Peers not managed by ansible from wireguard_unmanaged_peers
    [Peer]
    # client.example.com
    PublicKey = 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
    AllowedIPs = 10.0.0.3/32
```


**PublicKey** is required. **AllowedIPs** is effectively required but not a syntax error to be missing. **Endpoint** is not required, it's common to have it empty for roaming clients that will connect to this node.

githubixx:
@joneskoo To give you some feedback: I currently testing it. If it works out as intended I guess I'll merge it.

templates/wg.conf.j2
```diff
@@ -68,3 +68,26 @@ SaveConfig = true
     {% endif %}
   {% endif %}
 {% endfor %}
+{% if wireguard_unmanaged_peers is defined %}
+
+    # Peers not managed by ansible from wireguard_unmanaged_peers
+    {% for peer in wireguard_unmanaged_peers.keys() %}
+    [Peer]
+    # {{ peer}}
```
githubixx, line 76:
```suggestion
    # {{ peer }}
```

templates/wg.conf.j2
```diff
@@ -68,3 +68,26 @@ SaveConfig = true
...
+    {% if wireguard_unmanaged_peers[peer].public_key is defined %}
+    PublicKey = {{wireguard_unmanaged_peers[peer].public_key}}
```
githubixx, line 78:
```suggestion
    PublicKey = {{ wireguard_unmanaged_peers[peer].public_key|trim }}
```
This should be changed for L81,84,87 and 90 too.

githubixx (changes requested):
In general LGTM. Two comments inline.

But there is one thing that I'm not sure about if it should be left as is or changed. E.g. lets assume we've `host_vars/host1.yml` with:
```
wireguard_unmanaged_peers:
  client.example.com:
    public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
    allowed_ips: 10.8.0.218/32
```
and `group_vars/all.yml` with

```
wireguard_unmanaged_peers:
  client2.example.com:
    public_key: 6zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
    allowed_ips: 10.8.0.219/32
```
Personally I would assume that `client2.example.com` should appear in the WireGuard config of ALL hosts in the affected group. But that's not the case with the setup above. `host1` will only contain `client` and not `client2` peer config.

I think that's a perfectly valid use case to have clients that should be able to connect to all WireGuard hosts while others should be able to only connect to specific WireGuard hosts. WDYT?

j8r:
@githubixx The variable overriding is expected, that's how Ansible works.

One can create a variable, and merge it to the another one.

githubixx:
@j8r I know about the variable overriding but the question is if it should stay like this or if in this case the "global" variable and the host variable should be merged in general. As mentioned in my comment for me that's a perfectly valid use case.

README.md
```diff
@@ -95,6 +95,13 @@ wireguard_postup:
 wireguard_postdown:
   - ...
 wireguard_save_config: "true"
+wireguard_unmanaged_peers:
+  client.example.com:
+    public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
+    preshared_key: mfKkizs9J4eRFnyhaRvUKdFAxoGdDfcgMy4OhrbyjPE=
```
rubendibattista, line 101: Just a note. Maybe in the example use [`ansible-vault`](https://docs.ansible.com/ansible/latest/user_guide/vault.html) to provision the private key to promote good practices. :)

README.md
```diff
@@ -95,6 +95,13 @@ wireguard_postup:
 wireguard_postdown:
   - ...
 wireguard_save_config: "true"
+wireguard_unmanaged_peers:
+  client.example.com:
+    public_key: 5zsSBeZZ8P9pQaaJvY9RbELQulcwC5VBXaZ93egzOlI=
```
rubendibattista, line 100: I'm not very familiare with the pre-shared key config option, is that a secret? In this case maybe the README should promote good practices using `ansible-vault`

Legogris:
> @j8r I know about the variable overriding but the question is if it should stay like this or if in this case the "global" variable and the host variable should be merged in general. As mentioned in my comment for me that's a perfectly valid use case.


For the record, ansible can be configured with either behavior by the user: https://docs.ansible.com/ansible/2.3/intro_configuration.html#hash-behaviour

README.md
joneskoo: Do you have a change for me to include? Otherwise I'd leave psk out of the example.

It is secret but on its own does no good so I suppose it depends on your threat model whether you'd use Vault for that, and only for that?

templates/wg.conf.j2
```diff
@@ -68,3 +68,26 @@ SaveConfig = true
     {% endif %}
   {% endif %}
 {% endfor %}
+{% if wireguard_unmanaged_peers is defined %}
+
+    # Peers not managed by ansible from wireguard_unmanaged_peers
+    {% for peer in wireguard_unmanaged_peers.keys() %}
+    [Peer]
+    # {{ peer}}
+    {% if wireguard_unmanaged_peers[peer].public_key is defined %}
```
joneskoo, line 77: For public key can drop the if as it's required and not valid peer without it. 

I'd keep others optional although allowed ips is effectively required but not by syntax validation. Maybe someone would update allowed ips dynamically somehow externally. No harm in generating non-usable but valid config.

README.md
githubixx: You can read more about pre-shared keys here: https://wiki.archlinux.org/index.php/WireGuard#Key_generation (2nd half). In general I'd leave `ansible-vault` out. It's simply out of scope. Maybe mentioning `ansible-vault` is ok but that's it. If people want to use it they need to read the manual of `ansible-vault` anyways if they're not already familiar with it.

githubixx:
@Legogris I definitely don't want users to change their Ansible configuration just for that. That makes no sense and documentation also discourages to use that option. So `combine` filter is at least one way implement that behavior if desired. The point is I'm :100:  sure that someone will ask for that feature in the future :wink: That's why I want to discuss it now.

We can think about a variable to either enable or disable merging "global" and host `wireguard_unmanaged_peers` variables with default to "no". That's something I can think of currently. But if all people think that's not needed and current behavior is fine then that's ok for me too.

templates/wg.conf.j2
joneskoo: I don't see trim on other fields so why specifically for wireguard_unmanaged_peers?

joneskoo:
> We can think about a variable to either enable or disable merging "global" and host `wireguard_unmanaged_peers` variables with default to "no". That's something I can think of currently. But if all people think that's not needed and current behavior is fine then that's ok for me too.

My 2 cents. It's unlikely peers are defined both globally and per host, and if someone does that, that's a scenario that can be addressed when there's a concrete use case. I don't see breaking that behaviour in a future major version would be unthinkable if this turns out to be wrong.

Most important is to be able to define peers. I think they'll be special case and not that important. Besides it's always possible to use multiple playbooks to set up multiple tunnels for complex scenarios.

templates/wg.conf.j2
githubixx: The main idea here was that `trim` is a good idea in general but esp. if the variable was provided directly by the user and is not the result of some computation like the private/public keys that are automatically created e.g. I want to add this at other places too in the future where it makes sense.

joneskoo: I fixed the whitespace but didn't add trim - if it's added I think it makes sense to add on other variables too.

githubixx:
@joneskoo You're probably right. As I'm the only one who thought about that use case and since it should be able to add that later if needed we leave it as is and don't think about variable merging any further.

githubixx:
@joneskoo From my POV it's ready to go. If you don't have anything further to add I'd merge this.

joneskoo:
I don't have anything to add.

The failure when public_key is missing is not best ever but I think that's because of Ansible and not anything I added. Works OK in my brief testing. I have no idea how to make it better anyways.

```FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'public_key'"}```

githubixx (approved):
@joneskoo Thanks for that PR! We leave it as is now. Possible improvements can be done later if needed. Hopefully that makes life of some users easier now :smiley:

PR #64, opened 2020-09-01
README: add syntax highlighting language

joneskoo:
Github can render pretty syntax highlighting for YAML and ini code snippets if it knows the language. Set the language in the code block header.


githubixx (approved):
Thanks!

Issue #66, opened 2020-09-13
Combined effort with DebOps?


ypid:
Hey there, DebOps developer here. I found your role to be the best starting point to add WireGuard support to DebOps and would like to have your feedback on it.

For now, I see these viable options:

1. Make the role itself auto detect (Ansible local facts) when it runs against a DebOps managed host and behave in a compatibly way. Additionally, as DebOps is a comprehensive approach to configuration management, I would need to maintain a wrapper role in DebOps that manages the Firewall. An example how this can look like is DebOps’s tor relay role: https://github.com/debops/debops/tree/master/ansible/roles/tor and its corresponding playbook:
https://github.com/debops/debops/blob/master/ansible/debops-contrib-playbooks/service/tor.yml This is an intermidiate option but it would avoid the forking that I don’t like.

2. **Fork the role and make it a native DebOps role following https://docs.debops.org/en/master/dep/dep-0002.html**

   Edit: That is kinda unavoidable. I started this work at https://github.com/ypid/ansible-wireguard/tree/prepare-for-debops

3. We update the role and move it to DebOps :)

This is not urgent for me. I will open a improvement PR and a second one to add DebOps compatibility following option 1 and then you can have a look at the code.

Also note that DebOps has a role which solves similar problems of setting up a distributed public key system, https://docs.debops.org/en/master/ansible/roles/tinc/index.html. I will look into this older role again to compare your design with this one. Note that the tinc role has not be reworked since 2017 so it is not the most up-to-date one. There are more high-quality once if you want to check one out.

githubixx:
Thanks for your input and PR. Lets start the at the easy end :wink: After having a quick view on your draft PR https://github.com/githubixx/ansible-role-wireguard/pull/67 I think it's ok if it stays that way. Namespaces are ok for me. I've no problem with that kind of change.

In general this role was created to only serve one purpose and that was to enable secure communication of Kubernetes resources of my K8s cluster. I published it on Github for people who want to follow my little K8s installation blog series or have a similar need. I tried to make it as general as possible so that it can be used for other purposes too. But somewhere I need to draw I line because I can't maintain every use case. This is also the point where I tell people to fork the role. I've really no problem with that. That's how OpenSource works now and then, right? :wink:

[...]

PR #67, opened 2020-09-13
Improve coding style, spelling and Debian support

ypid:
~~WIP, I will do more testing.~~
Ref: #66


ypid:
I am not yet done with my rework and testing, but considering the high activity here, I will split my enhancements up and rather open additional PRs later. Feel free to review this one :)

tasks/setup-debian-vanilla.yml
```diff
@@ -1,37 +1,27 @@
...
-      - "linux-headers-{{ kernel_header_version }}"
+      - "linux-headers-{{ ('-cloud-' in ansible_kernel) | ternary(ansible_kernel,wireguard__fact_dpkg_arch.stdout) }}"
     state: present
 
-- name: (Debian) Install wireguard packages
+- name: (Debian) Install WireGuard packages
   apt:
     name:
-      - "wireguard-dkms"
-      - "wireguard-tools"
```
githubixx, line 34: What happens if the two packages are already present? Are they conflicting with `wireguard` package or will this just work because `wireguard` is a meta package?

meta/main.yml
```diff
@@ -11,7 +11,6 @@ galaxy_info:
     - focal
   - name: Debian
     versions:
-    - stretch
```
githubixx, line 14: If `stretch` is removed here then `... Debian 9 (Stretch), ...` should also be removed from README.

githubixx (commented):
LGTM basically. My tests worked fine. Two small comments inline. And a CHANGELOG entry should also be added. Even if it's mostly a "cosmetic" change because of the Debian package change. That's were I'm not sure about the consequences. If it's a breaking change then this change would be version `8.0.0`. If it's really not that dramatic then `7.3.0` would be good enough. 
We should also mention in the CHANGELOG that because of formatting changes in the WireGuard configuration file a `syncconf` will take place to prevent heart attacks for some people :wink: I guess you can just copy the comment which I added for the `7.2.0` release regarding this fact.

tasks/setup-debian-vanilla.yml
ypid: It is a small package with:

```
Depends: wireguard-modules (>= 0.0.20191219) | wireguard-dkms (>= 0.0.20200121-2), wireguard-tools (>= 1.0.20200827-1~bpo10+1)
```

So no conflict. The added benefit is what is written in `/usr/share/doc/wireguard/TODO.Debian`:

```
* wireguard metapackage should accept a 5.6 (or later) kernel instead
  of a module
```

So they will just update the `Depends:` line to accept a recent enough kernel.

meta/main.yml
ypid: Forgot about that, thanks. I fixed up the commit that to keep atomic commits.

I also rebased on latest master to resolve the conflicts.

ypid:
A hint when you merge. Feel free to merge this without squashing or a regular merge. This way, my signed commits will stay intact ;-)

ypid:
Done that. None of the changes should be breaking once.

CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
...
+
+  It does not serve any function anymore after support for module
+  reloading has been removed from the postinst script in 0.0.20200215-2 on
+  2020-02-24. A module update is properly signaled via
+  /run/reboot-required so that the admin can (automatically) schedule a
+  reboot when convenient. This will also be more in line with future Debian
+  releases because starting with Debian bullseye, the kernel ships the
+  module.
+- Add `ansible_managed` header to WireGuard configuration file (`wg0.conf` by default). This will most probably change the WireGuard configuration file but only the formatting. But since the Ansible registers this file as changed Ansible will sync/restart WireGuard service. For newer WireGuard versions (since Nov. 2019) this isn't a problem normally as `wg syncconf` command is used (also see `handlers/main.yml`). (contribution by @ypid)
+- Behind the sense coding style improvements and cleanup without user impact. (contribution by @ypid)
```
githubixx, line 18: Shouldn't this be `scenes` instead of `sense`?

ypid: Fixed, thanks!

PR #68, opened 2020-09-14
fix: create symlink for config to /etc/wireguard

kuriot:
A quick and dirty fix for wg-quick. 

Wg-quick searches for configs in `/etc/wireguard` and it cannot be changed, AFAIK. So, if for some host/group one changes `wireguard_remote_directory` variable, it will create a symlink of config to default location.


githubixx:
TBH I don't really understand the intension of this PR. Esp. the `/etc/hosts` thingy. Why is this needed? I've DNS entries for all hosts and that's good enough IMHO.

PR #69, opened 2020-09-16
Add wireguard_private_key variable

### LINKED ISSUE #65: Variables to set server public/private key
Maybe it cannot be considered as a best practice, but when having a lots of clients already configured, it is very impractical to bother everyone to change their configuration when we re-deploy our WireGuard instance.

ansible-vault is used, and the private key is occasionally rotated. Doing this tasks can be recommended in the documentation.

j8r:
Closes #65.

Setting the private key of the WireGuard VPN server allows to keep all the client configurations.
I didn't put the variable in `README.md` because it may be not recommended for everyone, and requires to encrypt the key in some way, but now it is possible to set it.

When testing my changes, I noticed the check mode was broken. It is now fixed in a commit.

I also changed a shell task to command, the later being recommended over shell.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ SPDX-License-Identifier: GPL-3.0-or-later
 Changelog
 ---------
 
+**7.6.0**
+- Add `wireguard_private_key` variable (contribution by @j8r)
+- Fix check mode for Debian (contribution by @j8r)
+
 **7.5.0**
 
 - `wireguard` package is now available for Ubuntu 18.04 in universe repository. Before that `ppa:wireguard/wireguard` was used but that one isn't available anymore. The install procedure for Ubuntu 18.04 and 20.04 is now the same as both can use `wireguard` metapackage now. The role takes care to remove `wireguard-dkms` package in favour of `wireguard` metapackage but it leaves the configuration file for `ppa:wireguard/wireguard` repository untouched. So it's up to you to remove that PPA. Either use `apt-add-repository --remove ppa:wireguard/wireguard` or remove the file manually at `/etc/apt/sources.list.d/` directory (you man need to run `apt-get update` afterwards).
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -34,3 +34,7 @@ wireguard_ubuntu_update_cache: "true"
 
 # Set package cache valid time
 wireguard_ubuntu_cache_valid_time: "3600"
+
+# This is sensitive: encrypt it with a tool like Ansible Vault.
+# If not set, a new one is generated on a blank configuration.
+# wireguard_private_key:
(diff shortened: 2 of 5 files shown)


githubixx (commented):
LGTM in general. But I don't see how this solves your issue https://github.com/githubixx/ansible-role-wireguard/issues/65 TBH. As it is implemented currently it will create `wg0.conf` with the supplied `wireguard_private_key` for the specified host if no `wg0.conf` exists. So far so good.

But you mentioned that you rotate the private key from time to time for your VPN server. But this won't happen when you update the value of `wireguard_private_key`. And even if it would update the private key, the public key also changes so you need to distribute this public key to your clients anyways. So IMHO this PR only solves a part of your issue.

Nevertheless please update CHANGELOG accordingly. This will be version `7.3.0` as I want to merge https://github.com/githubixx/ansible-role-wireguard/pull/61 first. So you might want to rebase after that merge. I guess there should be only little to no conflicts that needs to be solved.

j8r:
Usually we re-create the whole host, so won't be an issue. Nevertheless, you're right, can be fixed this by adding a condition to the `Set private key fact` task.

We do the private and public key generation locally. The later being already the case, as you said, needed by the "unmanaged" clients.

j8r:
@githubixx I think this PR will be ready to merge before #61, so I made it version 7.2.0 for now so you are free to do as you wish.

ypid (changes requested):
I find the way the private key is handled here very automation unfriendly and would not go this direction. I went into more detail here https://github.com/githubixx/ansible-role-wireguard/issues/65#issuecomment-694485885.

j8r:
@ypid I think this little PR is fine. It lets users do whatever they judge the best. For small setups, it is very fine to use a simple Ansible Vault. Of course, alternatively, other lookups are available like AWS Secret Manager.
Then, automation is done is done separately from the deployment, by some tool (can be Ansible too).

ypid:
> For small setups, it is very fine to use a simple Ansible Vault. Of course, alternatively, other lookups are available like AWS Secret Manager.

Probably worth adding an example that shows some integration into a proper external tool to handle the keys.

j8r:
@githubixx I rebased and test it again, is it ok for you?

j8r:
@githubixx Rebased, again :confused: 
Can you tell me what remains to do?

I don't think it is necessary to add an example to how using Ansible Vault or lookups.

githubixx:
@j8r Sorry, I needed to merge the Ubuntu 18 fix first as support for this OS was basically broken. I also needed to think about your PR a little bit first and I wanted to test it with various settings before I merge it. But it looks like that the PR is okay like it is. I'm still not sure if I shoot myself in the foot later but we'll see :wink: Thanks for your patience!

Issue #71, opened 2020-09-19
Machine readable copyright and license information, reuse.software


ypid:
Hi, what do you think of https://reuse.software/ ?

githubixx:
I don't want to put too much effort on licensing. I definitely won't sue somebody for forking this role and change it to meet their needs. As long as the changes are released as open source again I'm fine.

ypid:
reuse.software is DebOps standard. Would you be open to a PR which makes this role compliant with it by having the copyright info reversed from the Git history?

Issue #72, opened 2020-09-19
Exact license used (SPDX license indentifier): GPL-3.0-only or GPL-3.0-or-later?


ypid:
Which license do you mean exactly by "GNU GENERAL PUBLIC LICENSE Version 3" ? Is it [GPL-3.0-only](https://spdx.org/licenses/GPL-3.0-only.html) or [GPL-3.0-or-later](https://spdx.org/licenses/GPL-3.0-or-later.html)?

githubixx:
No idea what `-or-later` means :wink: There is nothing like that at https://www.gnu.org/licenses/. When I create a new repository at Github you can add a license and I choose always `GNU General Public License v3.0` (also see https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/licensing-a-repository). But they're all release on 29 June 2007 . I'm not a license expert and personally I don't care. The important thing for me is that software released as open source should stay open source. That's why I normally don't choose a BSD license.

ypid:
Default GPL license text is `-or-later` meaning that when a new version of the GPL gets released, users can apply it to your project without you having to review/approve this first. `-only` does not.

SPDX license identifier are the definitive way to identify licenses, also backed by the Linux Foundation.
https://www.linuxfoundation.org/blog/2018/08/solving-license-compliance-at-the-source-adding-spdx-license-ids/

This distinction is also important for #71.

githubixx:
In that case the `-or-later` option makes most sense to me. I guess FSF has good enough lawyers to not introduce crap in future versions of this license.

PR #76, opened 2020-09-26
REUSE Specification v3.0 and other minor stuff

ypid:
~~WARNING: This PR just assumes the default license `GPL-3.0-or-later` because no precise choice has been made yet (#72). I would propose to clarify #72 first before merging this.~~ Thanks for doing this!

I would say this PR does not need a changelog entry. If you disagree just ping me.

Closes: #71

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -1,4 +1,7 @@
 ---
+# Copyright (C) 2018-2020 Robert Wimmer
+# SPDX-License-Identifier: GPL-3.0-or-later
+
 - name: Gather instance facts
   setup:
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -2,6 +2,7 @@
 - name: Gather instance facts
   setup:
 
+# Installing WireGuard [[[1
 - include_tasks: "{{ item }}"
   with_first_found:
     - "setup-{{ ansible_distribution|lower }}-{{ ansible_distribution_major_version }}.yml"
@@ -23,6 +24,7 @@
     - wg-install
   when: not ansible_os_family == 'Darwin'
 
+# Key handling [[[1
 - name: Register if config/private key already exists on target host
   stat:
     path: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
@@ -89,6 +91,7 @@
   tags:
     - wg-config
 
+# Configuration [[[1
 - name: Create WireGuard configuration directory
   file:
     dest: "{{ wireguard_remote_directory }}"
@@ -116,6 +119,7 @@
   tags:
     - wg-config
 
+# Service [[[1
 - name: Start and enable WireGuard service
   service:
     name: "wg-quick@{{ wireguard_interface }}"
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -18,11 +18,11 @@
   modprobe:
     name: wireguard
     state: present
-  register: wireguard_module_enabled
-  until:  wireguard_module_enabled is succeeded
+  register: wireguard__register_module_enabled
+  until:  wireguard__register_module_enabled is succeeded
   retries: 10
   delay: 10
-  failed_when: wireguard_module_enabled is failure
+  failed_when: wireguard__register_module_enabled is failure
   tags:
     - wg-install
   when: not ansible_os_family == 'Darwin'
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
+# Copyright (C) 2018-2020 Robert Wimmer
+# SPDX-License-Identifier: GPL-3.0-or-later
+
 molecule/kvm/.vagrant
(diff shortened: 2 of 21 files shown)


githubixx (changes requested):
Please do not include email addresses without asking people for permission! If that is an requirement for this license thingy then please just close the PR. I'll definitely don't include email addresses. I'm not even sure if this violates GDPR.
Overall this process seems to be overly complicated to me and I'm not even sure if the copyrights are correct.

ypid:
I do care about the GDPR. I just included "public" Email addresses (either from the commit or GitHub profile), but I will recheck those. The authors already made their choice in my eyes and it is useful for an open source project to be able to reach the authors.

ypid:
All emails included are where already included in the commits, by the authors themselves:

```Shell
git shortlog -sne | cat
grep 'Copyright (C)' -R --no-filename | sort -u
```

ypid:
> I'm not even sure if the copyrights are correct.

I did it on a best effort basis. That was also done for the DebOps project once and I had to update a few of my copyrights afterwards. It is the best I can do. The people I listed should in all cases have a copyright on the file. It is very possible that people are missing from files. Everything else is documented by https://reuse.software/

j8r:
I doubt it's worth it to bother adding a license header on each file, for such project.
At the project's root, the license will be as obvious. Furthermore, that's not like we can enforce the license/copyright if someone just copy a whole file.

Also personally, I'd rather just use a simple permissive license like ISC or MIT.

ypid:
Have you read https://reuse.software/? It solves a problem and it is worth to just stick to it everywhere. It does not really hurt. One thing I really like is machine readable license information.

> Furthermore, that's not like we can enforce the license/copyright if someone just copy a whole file.

I agree that this cannot be prevented with such a header, but that is also not really my fear.

> I doubt it's worth it to bother adding a license header on each file, for such project.

Some things in life should just be dealt with, just to be on the safe side. I have seen enough projects that did re licensing of projects and the like. Not really something I as techy want to deal with and https://reuse.software/ looks like a proper spec to me.

Also see, #66, @githubixx and me agreed that me forking the role is the best way to move forward. I try to make all not invasive changes/improvements first so that the original can easily profit from it.

> Also personally, I'd rather just use a simple permissive license like ISC or MIT.

I also have my own believes here. But as license mixing gets messy and I often step down from my own favorite license (AGPL-3.0-only) when contributing to projects under different licenses just to keep it simple.

Fun fact (IANAL): It would not really matter when you would contribute code under ISC or MIT to this role because they are GPL compatible and the role as a whole would be GPL (dual licensing).

https://www.gnu.org/licenses/license-list.en.html
https://softwareengineering.stackexchange.com/questions/105912/can-you-change-code-distributed-under-the-mit-license-and-re-distribute-it-unde

Anyway.

tasks/main.yml
```diff
@@ -1,7 +1,11 @@
 ---
+# Copyright (C) 2018-2020 Robert Wimmer
+# SPDX-License-Identifier: GPL-3.0-or-later
+
 - name: Gather instance facts
   setup:
 
+# Installing WireGuard [[[1
```
githubixx, line 8: What is this `[[[...` useful for (here and elsewhere)?

githubixx (changes requested):
I generally agree with j8r's last statement and that basically matches with my first comment. But on the other hand it (hopefully) won't kill me adding this information and the PR is already there so we can do it. I still think it's overkill but YOLO :wink: 

The PR has currently conflicts that needs to be solved first. I added a molecule test yesterday. Maybe you can add the headers there too.

Also please remove the email addresses. Even if we think that the authors silently agreed that the email address can be included somewhere maybe that wasn't their intention. I personally absolutely hate it if such things happen. If I don't explicitly agree on something then someone can assume I didn't agreed at all. And I assume the same for other people.

And then one final question: What should happen to the license information at the end of the README: https://github.com/githubixx/ansible-role-wireguard#license ? Should this also be changed, removed or updated?

tasks/main.yml
ypid: That is why I make small commits to explain atomic changes: https://github.com/githubixx/ansible-role-wireguard/pull/76/commits/834967abfddedf1f9ed9b232c017b68f9321d510 Editor fold marks. The atomic commits now also easily allow me to remove this commit in case you don’t like it :)

githubixx: Well, no, I don't like them :wink: That's really confusing. There are thousands of editors out there and maybe one of them recognizes this by default. Such things really needs to be handled in the editor configuration if needed.

ypid:
Your feedback should be addressed I hope.

> Also please remove the email addresses. Even if we think that the authors silently agreed that the email address can be included somewhere maybe that wasn't their intention. I personally absolutely hate it if such things happen. If I don't explicitly agree on something then someone can assume I didn't agreed at all. And I assume the same for other people.

What I hate is when people don’t get the difference of Bcc vs. Cc :) Here on GitHub we are in the unprotected public. But I am fine with removing the email addresses. I rebased the commit so they are also not part of the history other than when the author has explicitly chosen to do this.

> Should this also be changed, removed or updated?

The statement is fine, but according to SPDX, there is also a more specific full name that I now changed to.

Issue #84, opened 2020-10-08
"Add WireGuard repository" fails on Ubuntu 18.04


l2fprod:
```
TASK [githubixx.ansible_role_wireguard : (Ubuntu) Add WireGuard repository (for Ubuntu < 19.10)] *******************************************************************************

fatal: [bastion]: FAILED! =>  {
    "changed": false,
    "invocation": {
        "module_args": {
            "codename": null,
            "filename": null,
            "install_python_apt": true,
            "mode": null,
            "repo": "ppa:wireguard/wireguard",
            "state": "present",
            "update_cache": true,
            "validate_certs": true
        }
    },
    "msg": "failed to fetch PPA information, error was: HTTP Error 404: Not Found"
}
```

It seems ppa:wireguard/wireguard no longer exists.

I tried the command manually:
```
$ add-apt-repository ppa:wireguard/wireguard
Cannot add PPA: 'ppa:~wireguard/ubuntu/wireguard'.
The team named '~wireguard' has no PPA named 'ubuntu/wireguard'
Please choose from the following available PPAs:
```

This is on
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 18.04.1 LTS
Release:	18.04
Codename:	bionic
```

githubixx:
Thanks for reporting! I guess your are right. Looks like that the WireGuard packages are now available in standard "universe" repo. Funny enough I tested this last week and it was still working. I'll have a look either today evening or during weekend and change it accordingly.

githubixx:
@l2fprod I just released 7.5.0 which should fix your problem. At least for my test instance it worked.

PR #85, opened 2020-10-11
Fix Ubuntu 18 install

githubixx:
`ppa:wireguard/wireguard` is gone. So it can't no longer be used for Ubuntu 18.04. Instead `wireguard` metapackage can be used now which is available in universe repository.


githubixx:
Closes https://github.com/githubixx/ansible-role-wireguard/issues/84

PR #88, opened 2020-11-23
Raspberry Pi: Use Backports instead of Debian Unstable

penguineer:
Use Debian backports instead of unstable to get wireguard. This is a more stable solution and has less impact on the system.

Unfortunately a reboot is still required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ SPDX-License-Identifier: GPL-3.0-or-later
 Changelog
 ---------
 
+**7.7.0**
+- Use wireguard packages from Debian Backports instead of Debian Sid, these packages are more suitable for a stable distribution and have less impact on the system. Packages from unstable must be removed manually (including kernel) to make the switch on an existing system. Upgrading the role has no effect other than adding Debian Backports to the Apt repositories.
+- Fix reboot mechanism in Raspbian role, now also works without `molly-guard`
+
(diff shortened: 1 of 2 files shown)


githubixx (commented):
Thanks for the PR! What happens to users that installed WireGuard with the former version of this role? Will it break their installations if they upgrade? Please updated CHANGELOG accordingly. Maybe add instructions how to upgrade from an older version of this role. If it's a breaking change please increase the major version number of the role accordingly. Otherwise increasing minor number should be good enough.

penguineer:
> What happens to users that installed WireGuard with the former version of this role? Will it break their installations if they upgrade? 

Good point.

My hypothesis is that nothing happens. There is a new repository and new versions, but they are older than the unstable packages. Also, the role will not update, only install, so a system that has been set up with unstable does not change.

An experiment has shown that my hypothesis is true: The only changes are the additional backports repository. After that, no changes and you can go forth and back between the versions.


In order to switch from unstable to backports, the unstable packages must be deinstalled (also kernel change), which I cannot do from the Ansible role, because I do not know if the user wanted to have unstable for some other reason.

penguineer:
I made an entry to the changelog and changed the version to **7.7.0**.
It might be 7.6.1, but I wanted to reflect that the system state will be different (Backports instead of unstable) when the new role is used.

githubixx (approved):
Thanks!

PR #93, opened 2021-02-06
Fix Fedora >= 33 support

wzzrd:
This commit fixes Fedora >= 33 support.

Fedora 32 still installs the copr repo and the dkms module through 
setup-fedora-32.yml I assume that is still necessary for Fedora 32, 
though I have no box to test it with.

If the user is on Fedora 33 or higher, the default setup-fedora.yml is
used, which no longer installs the copr repo, nor the dkms module since
neither are necessary anymore: the wireguard module has been merged
in the upstream kernel and wireguard-tools is in the base OS.


githubixx (approved):
Thanks!

PR #98, opened 2021-03-09
Amazon Linux 2 support

gavmain:
Hi there, I leaned heavily on this module to stand up a Wireguard instance on EC2, but found that it was lacking support for Amazon Linux 2. I'd be grateful if you would accept this PR. Thanks for your efforts with this role, it works like a charm.

My playbook which encapsulates this role: https://github.com/gavmain/bootstrap-wireguard-ec2

Thanks,
Gav

### DIFF
--- a/tasks/setup-amazon-2.yml
+++ b/tasks/setup-amazon-2.yml
@@ -1,0 +1,27 @@
+---
+# Copyright (C) 2020 Gav Main
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Install EPEL from Amazon Linux Extras
+  command: amazon-linux-extras install epel -y
+  args:
+    creates: /etc/yum.repos.d/epel.repo
+
+- name: (Amazon Linux 2) Add WireGuard COPR
+  yum_repository:
+    name: "jdoss-wireguard"
+    description: "Copr repo for WireGuard owned by jdoss"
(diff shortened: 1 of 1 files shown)


tasks/setup-amazon-2.yml
```diff
@@ -0,0 +1,27 @@
+---
+# Copyright (C) 2020 Gav Main
+# SPDX-License-Identifier: GPL-3.0-or-later
+
+- name: Install EPEL from Amazon Linux Extras
```
githubixx, line 5:
```suggestion
- name: (Amazon Linux 2) Install EPEL from Amazon Linux Extras
```

githubixx:
Hi Gav! Thanks for the PR! Before adding it I'd like to add a test for it. But I've a hard time to find an useful test image for my molecule tests: https://app.vagrantup.com/boxes/search?utf8=%E2%9C%93&sort=downloads&provider=libvirt&q=amazon
Most of them are outdated or not usable.

So I assume the value of the Ansible variable `ansible_distribution` is indeed `amazon-2`, right? Is this one based on Fedora Server? At least it looks a little bit like that...

Is the kernel of that distribution really that "old" that it still needs the WireGuard DKMS? Well "that old" is maybe a little bit harsh :wink: ... I know Debian 10 is using something like kernel v4.9. But for Ubuntu 18 it's possible to install newer kernels that already include the "wireguard" module e.g.

githubixx:
As I'm not really able to test this role with this distribution and no progress until a few month I'll close it.

Issue #101, opened 2021-04-25
hide peers without endpoint to enable routing via other wireguard host


Miroka96:
Hi!

In my current setup I use wireguard to manage very different hosts. Some are globally reachable servers, but most devices are behind firewalls. I need the possibility to access all devices from my own 'management' laptop, which is itself a device without global endpoint. Therefore, I have one dedicated wireguard server that serves as router between all NAT-ted devices. This 'router' has my whole IP range assigned as AllowedIPs. However, the routing does not work if the NAT-ted devices are set as peers without endpoints, because then routing requests (using ping) fail with `ping: sendmsg: Destination address required`.

Furthermore, devices without an endpoint also don't need to have a ListenPort as this port will never be used.

I suggest:
* to hide ListenPort, if the Endpoint is empty
* to hide the whole Peer section for devices whose Endpoint is empty

Best regards,
Miroka

githubixx:
Should be fixed with https://github.com/githubixx/ansible-role-wireguard/pull/213

PR #102, opened 2021-04-25
Hide peers with empty endpoints

Miroka96:
solves #101

### DIFF
--- a/templates/etc/wireguard/wg.conf.j2
+++ b/templates/etc/wireguard/wg.conf.j2
@@ -8,7 +8,9 @@
 # {{ inventory_hostname }}
 Address = {{ wireguard_address }}
 PrivateKey = {{ wireguard_private_key }}
+{% if wireguard_endpoint is defined and wireguard_endpoint != "" %}
 ListenPort = {{ wireguard_port }}
+{% endif %}
 {% if wireguard_dns is defined %}
 DNS = {{ wireguard_dns }}
 {% endif %}
@@ -45,7 +47,7 @@ PostDown = {{ wg_postdown }}
 SaveConfig = true
 {% endif %}
 {% for host in ansible_play_hosts %}
-{%   if host != inventory_hostname %}
+{%   if host != inventory_hostname and ((hostvars[host].wireguard_endpoint is defined and hostvars[host].wireguard_endpoint != "") or (wireguard_endpoint is defined and wireguard_endpoint != "")) %}
 
 [Peer]
 # {{ host }}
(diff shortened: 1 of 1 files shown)


githubixx:
Thanks for the PR! Sorry for not getting back earlier! While this PR works fine for my setup I'd like to setup an additional Molecule test scenario fist to make sure that it works as intended. I can't test every edge case but I'd at least have a better "feeling" before merging it :wink: I'm not sure if the current Molecule test scenario really covers this case at least a little bit.

jld3103:
It's good to know that you have to set `wireguard_port` to `""` otherwise it will still create then endpoint entries.

jld3103:
The same should be done for unmanaged peers:
```diff
diff --git a/templates/etc/wireguard/wg.conf.j2 b/templates/etc/wireguard/wg.conf.j2
index bc60039..ad758b3 100644
--- a/templates/etc/wireguard/wg.conf.j2
+++ b/templates/etc/wireguard/wg.conf.j2
@@ -88,6 +88,8 @@ Endpoint = {{host}}:{{wireguard_port}}
 
 # Peers not managed by Ansible from "wireguard_unmanaged_peers" variable
 {% for peer in wireguard_unmanaged_peers.keys() %}
+{%   if (wireguard_unmanaged_peers[peer].endpoint is defined and wireguard_unmanaged_peers[peer].endpoint != "") or (wireguard_endpoint is defined and wireguard_endpoint != "") %}
+
 [Peer]
 # {{ peer }}
 PublicKey = {{ wireguard_unmanaged_peers[peer].public_key }}
@@ -103,5 +105,6 @@ Endpoint = {{ wireguard_unmanaged_peers[peer].endpoint }}
 {%     if wireguard_unmanaged_peers[peer].persistent_keepalive is defined %}
 PersistentKeepalive = {{ wireguard_unmanaged_peers[peer].persistent_keepalive }}
 {%     endif %}
+{%   endif %}
 {%   endfor %}
 {% endif %}
```

almereyda:
I had merged this with master into https://github.com/libresh/ansible-role-wireguard, but cannot confirm that leaving the host variable out will also hide these clients.

christf:
I am in the same boat and tested your PR. It is working well for me.
@Miroka96  thank you for this patch. it did help me a lot!

Any chance to get this merged soon?

githubixx:
Hi! Can you please provide a test case that shows the problem? It would be very helpful to maybe make a copy of the existing [Molecule test](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/molecule.yml) with two or three VMs with some host variables that will cause the problem e.g. The test doesn't need to work as most people most probably don't have QEMU/KVM around to actually test it. I just want to have a few variables to play around. Thx!

varac:
Also looking forward for this, since the current behaviour breaks routing over a central server when two clients without endpoint try to connect to each other.

varac:
I tested this PR and it works nicely.

eugeneoden:
This PR helped me resolve routing issues when using empty endpoints for NAT'd peers without endpoints - seems safe to merge.

PR #106, opened 2021-05-12
Add flag to allow disabling deb repo management

tjend:
I manage my deb repos outside of this role, so this change allows this role to work with existing repos.

Also the wireguard package is all that is needed on my debian systems, so when using existing repos it also stops installing the kernel headers/dkms stuff. Let me know if this should be separated out into a separate pull request.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -25,6 +25,13 @@ wireguard_conf_group: "{{ 'root' if not ansible_os_family == 'Darwin' else 'whee
 wireguard_conf_mode: 0600
 
 
+######################################
+# Settings only relevant for Debian
+######################################
+
+# Set to "false" if apt repos are managed outside this role
+wireguard_debian_manage_deb_repo: "true"
+
 #######################################
 # Settings only relevant for Ubuntu
 #######################################
--- a/tasks/setup-debian-vanilla.yml
+++ b/tasks/setup-debian-vanilla.yml
@@ -10,12 +10,14 @@
     update_cache: yes
   tags:
     - wg-install
+  when: wireguard_debian_manage_deb_repo == "true"
 
 - name: (Debian) Install kernel headers for the currently running kernel to compile Wireguard with DKMS
   apt:
     name:
       - "linux-headers-{{ ansible_kernel }}"
     state: present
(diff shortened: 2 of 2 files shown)


defaults/main.yml
```diff
@@ -25,6 +25,13 @@ wireguard_conf_group: "{{ 'root' if not ansible_os_family == 'Darwin' else 'whee
 wireguard_conf_mode: 0600
 
 
+######################################
+# Settings only relevant for Debian
+######################################
+
+# Set to "false" if apt repos are managed outside this role
+wireguard_debian_manage_deb_repo: "true"
+
```
githubixx, line 34:
```suggestion
# Set to "no" if WireGuard binaries and tools are managed outside this role.
# In this case the role won't install the required WireGuard packages and it's
# up to you to install the required binaries before running this role!
# Currently this setting is only respected on hosts running Debian OS!
wireguard_manage_package: "yes"

```
I'd like to keep this generic. I can imagine that this is also helpful for people running other Linux distributions. I'm totally happy with the variable name but couldn't find anything better :wink:

tasks/setup-debian-vanilla.yml
```diff
@@ -10,12 +10,14 @@
     update_cache: yes
   tags:
     - wg-install
+  when: wireguard_debian_manage_deb_repo == "true"
```
githubixx, line 13: Please put https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-debian-vanilla.yml#L6-L31 into a `block`. This avoids all that `when` clauses. The `(Debian) Get architecture` task is also only needed if the Linux headers are needed. An example of such a `block` is here: https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-ubuntu.yml#L12-L24 in case you need the syntax.

githubixx (changes requested):
Thanks for the PR! You might need to rebase on master as I merged another PR which also touches `defaults/main.yml`. Please also update `README` accordingly and add line to `CHANGELOG` accordingly. I guess this will become version `7.12.0` then.

githubixx:
Implemented via https://github.com/githubixx/ansible-role-wireguard/commit/314fec524804d86d3512c131137335cdd3e17f42

Issue #81, opened 2020-09-27
Use no_log for tasks that might output sensitive information in --diff mode


ypid:
https://docs.ansible.com/ansible/latest/reference_appendices/logging.html#protecting-sensitive-data-with-no-log

From discussion in #78. Tracked here separately.
Fixed in: https://github.com/ypid/ansible-wireguard/tree/prepare-for-debops

githubixx:
I guess introducing the `no_log: True` attribute isn't that invasive if it's added to some tasks. But honestly it's not really the task of this role to protect people from shipping log output/secrets to a central logging system e.g. I guess people with strong requirements in this regards already have adjusted a fork of this role accordingly and will use `ansible-vault`, Ansible Tower or something like that.

ypid:
I thought you support the GDPR. "Privacy by default" :)

> I guess

You know how it is. People pull down your role, go like: "oh look, it works". A year later, it runs as part of their site playbook. Someone comes and wants to detect config drift, another departments sets up Elastic for logging. By the time they all forgot that this role might output sensitive data to the Elastic to which others might have access to.

Our IT is to complex to take guesses. We need strong foundations.

PR #110, opened 2021-07-04
Fix tag "wg-install" & Add no_log

### LINKED ISSUE #109: Tag wg-install is not applied properly
Hi altogether,

first of all: Thank you for this role - helped me rolling out wireguard on my hosts.

As I do have some hosts, that do run custom distros, the install tasks do not cover those.
No problem here - I think I've build myself some edge cases there.

But because of this, I've tried using the existing tag "wg-install" to skip any installation tasks and found it not working.
After a short inspection of the tasks, I've found a solution, that I'll provide in a PR later today.

This issue is more for documentation purposes than asking for a fix. :)

moonrail:
This PR fixes #109 & #81 

Background to the change on the tag:
Tags are evaluated at the start of a play, so any include, that is evaluated at runtime, will not result in having any tag-functionality applied to included tasks/roles, unless tags are applied via the parameter "apply".
https://docs.ansible.com/ansible/latest/user_guide/playbooks_tags.html#tag-inheritance-for-includes-blocks-and-the-apply-keyword
By adding "tags" to the main.yml-include_tasks-Task, this task is respected, if "wg-install" is given via "skip-tags" or "tags".
By adding "apply"-"tags", the defined tag "wg-install" ist applied to all included tasks and all included tasks are handled, when given the tag "wg-install" via "tags"-option.

As for no_log:
Default verbosity of 0 or slight increases up to 2 will now not print any private keys to output.
If anyone desires to see the output of these tasks, the verbosity can be increased to 3 or more to disable no_log.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -6,13 +6,19 @@
   setup:
 
 # Installing WireGuard [[[1
-- include_tasks: "{{ item }}"
+- include_tasks:
+    file: "{{ item }}"
+    apply:
+      tags:
+        - wg-install
   with_first_found:
     - "setup-{{ ansible_distribution|lower }}-{{ ansible_distribution_major_version }}.yml"
     - "setup-{{ ansible_distribution|lower }}-{{ ansible_distribution_version }}.yml"
     - "setup-{{ ansible_distribution|lower }}-{{ ansible_distribution_release }}.yml"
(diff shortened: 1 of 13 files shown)


githubixx (approved):
Thanks for your contribution! LGTM.

PR #112, opened 2021-07-09
Remove unnecessary and buggy check on wireguard_port

abelfodil:
This check is preventing proper port from being set (it defaults to the port of the current host instead of the targeted host).

Without this patch, [my use case](https://github.com/arch-anes/self-hosted-services/blob/f733d069b6124ac886b5d6c92106732d7380d1c8/roles/vpn/tasks/main.yml#L39) doesn't work, i.e. the value isn't picked up.


abelfodil:
@githubixx could you please review?

almereyda:
I can confirm that this patch works against current master to allow different hosts to use different ports.

https://github.com/libresh/ansible-role-wireguard

githubixx:
If you define the variable value as integer it should just work e.g. `wireguard_port: 51821. In the molecule test https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/molecule.yml#L158 the `wireguard_port` variable values are all integers. But I agree that it's not really needed. So I'll merge the PR. Thanks!

githubixx (approved):

PR #117, opened 2021-08-26
Add Debian 11 and Raspbian Buster to tested systems

penguineer:
I have tested the role on these two additional systems and they work perfectly fine.


penguineer:
I can confirm that a current raspbian image (`2021-05-07-raspios-buster-armhf-lite`) does not need any additional kernel headers or the `wireguard-dkms` package.

The Raspberry Pi installation task can be reduced to installing the packages `wireguard` and `wireguard-dkms`.

PR #119, opened 2021-08-30
Specify Raspbian playbook for Buster and below

penguineer:
Kernel Updates are no longer needed as Raspbian now comes with kernel 5.10 and directly supports wiregard.

Keep the distinction to enable other steps if they may come up at a later time.


j8r:
Is it still useful to differentiate Raspbian and Debian then?

githubixx:
This mostly depends on what's the value of `ansible_lsb.major_release` is as [tasks/setup-debian-vanilla.yml](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-debian-vanilla.yml#L33-L34) differentiates here. If the value is `11` @j8r is probably right and we can get rid of `tasks/setup-debian-raspbian.yml`.

penguineer:
I agree that the distinction is not needed right now. 

However, it seems that Debian and Raspbian go different ways sometimes. That's why I left it in.

(i.e. I would not have added it right now, but since it was there I didn't remove it, thinking that I might come in handy at a later point.)

I leave the decision to the maintainer.

j8r:
The best is not to delete this file, and use the Debian file only for the recent Raspbian releases.

penguineer:
> The best is not to delete this file, and use the Debian file only for the recent Raspbian releases.

Yes, to keep the role viable for older systems. I'll check if I have a test setup for that.

j8r:
Please do the opposite, the included tasks from `setup-debian-raspbian.yml` are defined in https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-debian.yml.

Just add another version check to prevent including the raspbian file.

githubixx:
Shouldn't it be good enough to just keep everything as is (to also support older versions - at least for a while) and just extend the `when` clause [here](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-debian.yml#L6-L11) like so:

```yaml
when:
  - ansible_lsb.id is defined
  - ansible_lsb.id == "Raspbian"
  - ansible_lsb.major_release is version('11', '<')
 ```

penguineer:
Yes, we can do that. I will make the change as soon as I have the time (most likely this weekend).

penguineer:
> Yes, we can do that. I will make the change as soon as I have the time (most likely this weekend).

Narrator: he didn't.

> Shouldn't it be good enough to just keep everything as is (to also support older versions - at least for a while) and just extend the `when` clause [here](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-debian.yml#L6-L11) like so:
> 
> ```yaml
> when:
>   - ansible_lsb.id is defined
>   - ansible_lsb.id == "Raspbian"
>   - ansible_lsb.major_release is version('11', '<')
> ```

I thought about this for a while.
a) Yes, it solves the problem.
b) It makes the role even harder to read: There is a raspberry pi part in the role and whoever wants to to something RPi specific can add it there - and then scratch their head why it isn't applied.

I will do the proposed addition and rename the role to `raspbian-buster` to avoid confusion.

penguineer:
(Didn't have a chance to test it yet.)

l0calhost:
@penguineer: I can confirm your fixes work for Raspberry OS 11.1. I tried this manually after the following task failed:

`TASK [ansible-role-wireguard : (Raspbian) Add Debian repository keys] ***************************************************************
[WARNING]: The loop variable 'item' is already in use. You should set the `loop_var` value in the `loop_control` option for the task
to something else to avoid variable collisions and unexpected behavior.
failed: [backupbox] (item=04EE7237B7D453EC) => {"ansible_loop_var": "item", "changed": false, "cmd": "/usr/bin/apt-key adv --no-tty --keyserver keyserver.ubuntu.com --recv 04EE7237B7D453EC", "forced_environment": {"LANG": "C", "LC_ALL": "C", "LC_MESSAGES": "C"}, "item": "04EE7237B7D453EC", "msg": "Error fetching key 04EE7237B7D453EC from keyserver: keyserver.ubuntu.com", "rc": 2, "stderr": "Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).\ngpg: keyserver receive failed: No route to host\n", "stderr_lines": ["Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).", "gpg: keyserver receive failed: No route to host"], "stdout": "Executing: /tmp/apt-key-gpghome.s1oFkiRuas/gpg.1.sh --no-tty --keyserver keyserver.ubuntu.com --recv 04EE7237B7D453EC\n", "stdout_lines": ["Executing: /tmp/apt-key-gpghome.s1oFkiRuas/gpg.1.sh --no-tty --keyserver keyserver.ubuntu.com --recv 04EE7237B7D453EC"]}
failed: [backupbox] (item=648ACFD622F3D138) => {"ansible_loop_var": "item", "changed": false, "cmd": "/usr/bin/apt-key adv --no-tty --keyserver keyserver.ubuntu.com --recv 648ACFD622F3D138", "forced_environment": {"LANG": "C", "LC_ALL": "C", "LC_MESSAGES": "C"}, "item": "648ACFD622F3D138", "msg": "Error fetching key 648ACFD622F3D138 from keyserver: keyserver.ubuntu.com", "rc": 2, "stderr": "Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).\ngpg: keyserver receive failed: No route to host\n", "stderr_lines": ["Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).", "gpg: keyserver receive failed: No route to host"], "stdout": "Executing: /tmp/apt-key-gpghome.r3C12esGR2/gpg.1.sh --no-tty --keyserver keyserver.ubuntu.com --recv 648ACFD622F3D138\n", "stdout_lines": ["Executing: /tmp/apt-key-gpghome.r3C12esGR2/gpg.1.sh --no-tty --keyserver keyserver.ubuntu.com --recv 648ACFD622F3D138"]}`

penguineer:
In my view the PR is ready for merge.

Sorry for the long delay and thank you for your patience!

Issue #120, opened 2021-08-31
setting wg0.conf in only 1 host from group


emper0r:
Hi again , i need set the configuration in only 1 host from a group

i have
```
[aws_fws]
fw1.aws.local
fw2.aws.local
```
the variables are in the file 
```
inventory/host_var/fw1/wireguard.yml
```

if I do 

```
ansible-playbook wireguard.yml --limit="fw1.aws.local" 
```
everything is OK in 1 host... BUT

if i do in my playbook the same thing 

```
- name: Configure Wireguard
  hosts: aws_fws
  gather_facts: yes

  vars:

  vars_files:

  roles:
    - role: wireguard
      when: inventory_hostname == 'fw1.aws.local'
```

i receive this error...

```
The full traceback is:
Traceback (most recent call last):
  File "/usr/local/bin/python_virtualenvs/ansible-4.3.0-devops_env9/lib/python3.8/site-packages/ansible/template/__init__.py", line 1100, in do_template
    res = j2_concat(rf)
  File "<template>", line 119, in root
  File "/usr/local/bin/python_virtualenvs/ansible-4.3.0-devops_env9/lib/python3.8/site-packages/jinja2/runtime.py", line 903, in _fail_with_undefined_error
    raise self._undefined_exception(self._undefined_message)
jinja2.exceptions.UndefinedError: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'wireguard__fact_public_key'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/python_virtualenvs/ansible-4.3.0-devops_env9/lib/python3.8/site-packages/ansible/plugins/action/template.py", line 146, in run
    resultant = templar.do_template(template_data, preserve_trailing_newlines=True, escape_backslashes=False)
  File "/usr/local/bin/python_virtualenvs/ansible-4.3.0-devops_env9/lib/python3.8/site-packages/ansible/template/__init__.py", line 1137, in do_template
    raise AnsibleUndefinedVariable(e)
ansible.errors.AnsibleUndefinedVariable: 'ansible.vars.hostvars.HostVarsVars object' has no attribute 'wireguard__fact_public_key'
fatal: [fw1.aws.local]: FAILED! => changed=false
  msg: 'AnsibleUndefinedVariable: ''ansible.vars.hostvars.HostVarsVars object'' has no attribute ''wireguard__fact_public_key'''

```

fw1.yml 
```
---
[...]

PR #121, opened 2021-09-01
* New: Set managed peer optional to install single server or multiserver on mesh mode

emper0r:
Hi about my issue: #120 we can found a solution to implement in our case use.

this role by default create mesh wireguard connection witch is very useful and as plus we add a new variable condition to can install as a single server or multiple single server in the same ansible_group, 

this new variable is defined by default to work backward compatible role.

thanks for your effort.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -6,6 +6,8 @@
 # General settings
 #######################################
 
+wireguard_managed_peers_enabled: True
+
 # Directory to store WireGuard configuration on the remote hosts
 wireguard_remote_directory: "{{ '/etc/wireguard' if not ansible_os_family == 'Darwin' else '/opt/local/etc/wireguard' }}"
 
--- a/templates/etc/wireguard/wg.conf.j2
+++ b/templates/etc/wireguard/wg.conf.j2
@@ -44,6 +44,7 @@ PostDown = {{ wg_postdown }}
 {% if wireguard_save_config is defined %}
 SaveConfig = true
 {% endif %}
+{% if wireguard_managed_peers_enabled %}
 {% for host in ansible_play_hosts %}
 {%   if host != inventory_hostname %}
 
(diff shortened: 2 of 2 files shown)


githubixx:
Possible solution mentioned in https://github.com/githubixx/ansible-role-wireguard/issues/120#issuecomment-976378382

stejoo:
Hi! 
First of all thank your for your efforts and resulting role. It's awesome. :)
Could it possible be even better by revisiting this issue/merge request?

I have been considering something similar to this pull request. Or defining a variable called `wireguard_managed_peers` that would be used instead of `ansible_play_hosts`. To not modify the behaviour of the role `wireguard_managed_peers` would default to `{{ ansible_play_hosts }}` for example, and allow somebody (or me) to specify a list of host names or a group name(s) to add as peers. I have not built anything like that yet because I thought "let see if somebody had the same idea before" and that's how I ended up here.

Creating a mesh as you are is fabulous. And @emper0r seems to have had an issue with a single client that could be resolved by putting it in a group of it's own. But I'm in a different boat... where I have nodes in the field, who should not communicate with (or even know about) each other. These field nodes do communicate with three reachable nodes that provide a way in to our backend. So I would like to add just these three added as peers, instead of the sisters nodes out in the field. While preserving a complete inventory of them, where I do have them neatly grouped, but of course all are in `ansible_play_hosts` when I try to push a change across to all of them.

Even completely disabling the addition of Ansible managed peers, as would be possible by setting `wireguard_managed_peers_enabled` to `false`, would work for me. I can manage three hosts in `wireguard_unmanaged_peers` just fine, as those are quite static anyway.

In the mean time I will probably have a go at my idea of replacing `ansible_play_hosts` with something called `wireguard_managed_peers` for now. 

What do you think of my conundrum? And would you be interesting in something like the above by any chance? :smile:

PR #123, opened 2021-09-28
Add Rocky Linux/AlmaLinux support + Molecule tests

githubixx:


gitouche-sur-osm:
Works flawlessly, thank you very much!

PR #129, opened 2021-10-05
feat: Update CentOS 7 to use signed kernel-plus module

john-p-potter:
Updated the CentOS 7 tasks to use kernel-plus and the built-in signed module for WireGuard instead of the standard kernel and ELRepo's module to better support hardening that requires signed kernel modules.  The negative is updating the kernel obviously requires a reboot, where using the standard kernel as previous does not.


githubixx (commented):
Thanks for the PR! I don't want to change the default behavior. I've no idea how many rely on the current installation method and changing that could break peoples setup. So this would be an optional installation method.

So first step probably would be to introduce a variable like `wireguard_centos7_installation_method` with the value `standard` (for `standard kernel`). Additional value could be `kernel-plus` e.g. Maybe it makes also sense to let the users decide if they want to reboot and how long they want to wait for the reboot. For this  `wireguard_centos7_kernel_plus_reboot` (with the values `true` or `false`) and `wireguard_centos7_kernel_plus_reboot_timeout` variables might be needed (that's the `reboot_timeout` parameter of the `reboot` module).

Currently this role expects Ansible `2.5` or later. If you introduce the `reboot` module this needs to be [raised](https://github.com/githubixx/ansible-role-wireguard/blob/master/meta/main.yml#L8) to at least `2.7`. I wanted to increase this to `2.9` anyways sooner or later so that would be ok.

Next step most probably would be to introduce some `block:`'s like we have it for Ubuntu  (https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-ubuntu.yml#L10-L21 e.g.) to figure out if `standard` or `kernel-plus` installation mode was requested and act accordingly.

That would be my suggestion. I hope I didn't forget something :wink:

john-p-potter:
Thanks for the feedback, I agree that is a good/better way to go - not changing the default behavior.  I'll update and test it out with your recommendations.  As for the reboot, do you want that moved to the centos7 setup task or keep it in handlers?

john-p-potter:
Updated per changes requested.  Good catch on the reboot check not being idempotent.

PR #132, opened 2021-10-10
Add  to allow host specific AllowedIPs configuration

tobias-richter:
At the moment it is not possible to set the `AllowedIPs` on host base.
In my setup this is required because one peer is part of two subnets.

```
       peer_a <->
                  peer_b
backup_peer_a <-> 
```
peer_a has the subnets 192.168.0.0/24
backup_peer_is connected to the subnets 192.168.0.0/24 and 192.168.1.0/24 (Backup Line)
peer_bhas the subnets 192.168.2.0/24

Assuming these are the configs:

`host_vars/peer_a.yaml`:
```yaml
wireguard_address: 10.8.0.1/24
wireguard_allowed_ips: 10.8.0.1/32, 192.168.0.0/24
```
`host_vars/backup_peer_a.yaml`:
```yaml
wireguard_address: 10.8.0.2/24
wireguard_allowed_ips: 10.8.0.2/32, 192.168.1.0/24
```

**Problem:**

`backup_peer_a ` will get a peer for `peer_a ` with `AllowedIPs=10.8.0.1/32, 192.168.0.0/24`.
Since `backup_peer_a ` is also directly connected to `192.168.0.0/24` this will lead to routing issues and the host may not be reachable from hosts within `192.168.0.0/24`.

So I introduced `wireguard_allowed_ips_override` to able to override `AllowedIPs` based on `inventory_hostname`.

In this example we would add the override in the `host_vars/peer_a.yaml`:
```yaml
wireguard_address: 10.8.0.1/24
wireguard_allowed_ips: 10.8.0.1/32, 192.168.0.0/24

wireguard_allowed_ips_override:
  "backup_peer_a": 10.8.0.1/32
```
So when the managed peer config of `peer_a` is rendered on `backup_peer_a` the value from the `wireguard_allowed_ips_override` is taken.

@githubixx i hope you could follow me so far :) Another solution to solve this problem would be to make the template path adjustable to be able to work with a custom template.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ SPDX-License-Identifier: GPL-3.0-or-later
 Changelog
 ---------
 
+**8.5.0**
+
+- add `wireguard_allowed_ips_override` to allow host specific AllowedIPs configuration (contribution by @tobias-richter)
+
 **8.4.0**
 
 - add support for installing wireguard in pve lxc guest (contribution by @tobias-richter)
--- a/README.md
+++ b/README.md
@@ -112,10 +112,20 @@ Endpoint = controller01.p.domain.tld:51820
 
 Now this is basically the same as above BUT now the config says: I want to route EVERY traffic originating from my workstation to the endpoint `controller01.p.domain.tld:51820`. If that endpoint can handle the traffic is of course another thing and it's up to you how you configure the endpoint routing ;-)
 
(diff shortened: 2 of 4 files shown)


tobias-richter:
After rethinking the solution I think i will go for the customizable template option. If you think the functionality of this PR may be useful let me know.

githubixx:
Maybe it makes sense to allow users to include their own `wg.conf.j2` template similar to https://github.com/githubixx/ansible-role-traefik-kubernetes/blob/master/tasks/install.yml#L15-L30. I also thought about that but so far it wasn't requested.

PR #133, opened 2021-10-11
add support to customize config template path

tobias-richter:
This PR enables the user to use his own template in order to implement customizations.


githubixx:
Thanks for the PR! But I came to the conclusion if someone wants to do that it's actually better just to fork the repository as basically the whole "logic" is in that file. If the provided version doesn't work for someone a fork is definitely the best option.

Issue #135, opened 2021-12-14
Execute lifecycle hooks upon reconfiguration to maintain consistency


8ware:
Managing `iptables` rules is a common example for using `wg-quick`'s lifecycle hooks `PreUp`, `PostUp`, `PreDown` and `PostDown`. However, when changing those rules and deploying this role again they're not executed since just the plain `wg` configuration is sync'ed. What's worse, after updating the configuration WireGuard can't be even restarted with `wg-quick` since deleting a non-existing `iptables` rule results in an error which in turn aborts the shutdown halfway through.

Suppose the following `PostUp` and `PreDown` rules are deployed:

```yaml
wireguard_postup:
  - iptables -A FORWARD -i %i -j ACCEPT
  - iptables -A FORWARD -o %i -j ACCEPT
  - iptables -t nat -A POSTROUTING -i %i -o eth0 -j MASQUERADE
# Generate inverse iptables commands
wireguard_predown: "{{ wireguard_postup | reverse | replace('-A', '-D') }}"
```

When I add a new rule (e. g. `iptables -A INPUT -i % -j REJECT`) to `wireguard_postup` and deploy the role again the configuration file contains two new lines: one `PostUp` and one `PreDown` entry (containing the new rule and its counterpart). When I try to restart WireGuard the counterpart of the new rule is already executed leading to an error.

On the otherhand, when I remove an `iptables` rule from `wireguard_postup` its counterpart is removed as well and won't be executed when restarting WireGuard leaving the system in an unclean state.

One way of resolving this issue is to shutdown the interface before updating the configuration file and starting it up again after the update. This comes with several challenges, though:

* Active connections will be interrupted (which might not be that big of a problem in a private network environment)
* To achieve idempotence we don't want to shutdown WireGuard if the configuration didn't change
* In order to run the `*Down` rules of the old (currently active) configuration we cannot update the configuration before shutting down the interface

Since there might be even more side effects I don't anticipate I'd like to start a discussion on that topic. Thank you in advance.

8ware:
Here are some approaches to the above mentioned challenges:

* We shutdown WireGuard only if it's necessary, i. e. if a `*Down` rule has changed; in that case we probably want to restart WireGuard anyway
* We use `check_mode` to determine whether the configuration has changed and `diff` to examine what in particular
* We introduce a shutdown handler and flush it before actually updating the configuration

Here is a code snippet to illustrate my approach:

```yaml
- ansible.builtin.template:
[...]

PR #138, opened 2022-01-04
General improvements

8ware:
* The resulting order of WireGuard's hooks should coincide with its lifecycle
* Using FQCNs is recommended by the official Ansible documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,12 @@ SPDX-License-Identifier: GPL-3.0-or-later
 Changelog
 ---------
 
+**9.0.0**
+
+- set minimally required Ansible version to 2.9
+- fully qualify modules names (requires Ansible >= 2.9)
+- rearrange hooks to match lifecycle order
+
 **8.4.0**
 
 - add support for installing wireguard in pve lxc guest (contribution by @tobias-richter)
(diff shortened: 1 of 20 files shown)


8ware:
FYI: This is a preliminary pull request. I implemented the actual solution for issue #135 in branch [8ware:consistent-hook-execution](/8ware/ansible-role-wireguard/commits/consistent-hook-execution).

githubixx (commented):
Thanks for the PR! That's something I wanted to do anyways sooner or later :wink: I'm not 100% sure but I guess the FQDN commands were introduced together with Ansible collections in Ansible 2.9. But I was not able to verify that. But https://docs.ansible.com/ansible/latest/reference_appendices/faq.html mentions that Ansible 2.9 introduced collections and I think that was also the introduction of FQDN commands. That would also mean that https://github.com/githubixx/ansible-role-wireguard/blob/master/meta/main.yml#L8 needs to be changed end set to `2.9`. And this in turn means that the new version of this role would be `9.0.0` as this is a breaking change at least for some people. So this also something the `CHANGELOG` should mention.

I haven't tested the change yet but having a quick look it's LGTM.

8ware:
Thanks for your remarks. I updated the role metadata and adjusted the new release version to `9.0.0` as suggested.

githubixx (approved):
LGTM. Thanks!

PR #141, opened 2022-01-16
Consistent lifecycle hook execution

8ware:
This change extends WireGuard's reconfiguration by executing its lifecycle hooks if they're changed (see #135). WireGuard will be shutdown *before* applying those changes to a) execute the `PreDown` and `PostDown` hooks of the current and later previous configuration version and b) execute the `PreUp` and `PostUp` hooks *after* the might have changed.

In order to avoid unnecessary shutdowns of WireGuard and thus connection interrupts WireGuard is only shut down if any of those hooks are getting changed.

Note that this may introduce logging of sensitive information as `ansible.utils.fact_diff` does not work if the `no_log` property is used.


tasks/main.yml
```diff
@@ -122,6 +122,53 @@
...
+  ansible.builtin.template:
+    src: etc/wireguard/wg.conf.j2
+    dest: "{{ wireguard_remote_directory }}/{{ wireguard_interface }}.conf"
+    owner: "{{ wireguard_conf_owner }}"
+    group: "{{ wireguard_conf_group }}"
+    mode: "{{ wireguard_conf_mode }}"
+  check_mode: yes
+  diff: yes
+  changed_when: no
+  register: wg_config
```
githubixx, line 135: Internal variables start with `wireguard__`:
```suggestion
  register: wireguard__register_configuration
```

tasks/main.yml
```diff
@@ -122,6 +122,53 @@
...
+  tags:
+    - wg-config
+
+- name: Examine configuration changes | Execute lifecycle hooks if necessary
+  block:
+    - name: Compute differences between present and future configuration
+      ansible.utils.fact_diff:
+        before: "{{ wg_config.diff[0].before }}"
+        after: "{{ wg_config.diff[0].after }}"
+      register: wg_config_diff
```
githubixx, line 145:
```suggestion
      register: wireguard__register_configuration_diff
```

tasks/main.yml
```diff
@@ -122,6 +122,53 @@
...
+      changed_when: no
+
+    - name: Check if any lifecycle hooks have changed
+      ansible.builtin.set_fact:
+        wg_hooks_changed: "{{
```
githubixx, line 150:
```suggestion
        wireguard__fact_hooks_changed: "{{
```

githubixx (commented):
Thanks for the PR! I've to test it first and think a little bit about it. But after a quick look right now the WireGuard configuration file (like `/etc/wireguard/wg0.conf`) would be rendered twice, right? Because in [line 172](https://github.com/githubixx/ansible-role-wireguard/pull/141/files#diff-3d0ff1709ca48add100327bb2a468e6c508fb92a159c64c4f99ad1df89d9bddeR172) the old template task is still there.

Also for some people a shutdown might not be needed. It's possible to execute a Shell or Python script with the hooks and do whatever you want. So I'd prefer to be able to disable this behavior or even better disable the behavior by default. In that case the version number can be left at `9.1.0`. It'd be a feature in that case that doesn't change the previous behavior of the role. I don't have a good idea for a variable right now. Maybe `wireguard_interface_restart_on_hook_change: false`.

githubixx:
There is now a `wireguard_interface_restart` variable. I think that good enough for most use cases.

PR #143, opened 2022-01-19
Examples should use public not private keys

whalecoiner:
Perhaps I've misunderstood Wireguard, but shouldn't the `peers` in all the examples be using `PublicKey` rather than `PrivateKey`?


githubixx (approved):
Oh how funny is that... :facepalm:  Nobody (including myself) ever recognized this since years. If you read a text too often you don't see the obvious bugs anymore :wink: Thanks for bringing this on the table and providing a PR!

sonniesedge:
Haha glad to have helped!

PR #146, opened 2022-02-05
Add non-standard kernel installation for RockyLinux 8

gitouche-sur-osm:
Hello, I've been toying with RockyLinux on Raspberry 3 (`aarch64`) and needed the [non-standard kernel](https://www.wireguard.com/install/) installation, as documented by wireguard.

Although my use case is very specific, it should work with any custom kernel on `aarch64`, `ppc64le` and `x86_64`.  
It might also work on AlmaLinux (not tested)

I expanded on the logic you had for CentOS7 setup with a new variable : wireguard_rockylinux8_installation_method, defaulting to the classic EPEL/Elrepo installation process.

### DIFF
--- a/tasks/setup-rocky-8.yml
+++ b/tasks/setup-rocky-8.yml
@@ -24,8 +24,7 @@
           - "wireguard-tools"
         state: present
   when:
-    - (wireguard_rockylinux8_installation_method == "standard" and ansible_architecture == "x86_64" and ansible_kernel is match("4\.18\.0-.*x86_64")) or
-      (wireguard_rockylinux8_installation_method == "kmod")
+    - wireguard_rockylinux8_installation_method == "standard"
 
 - name: (Rocky Linux 8) Tasks for non-standard kernel
   block:
@@ -54,5 +53,5 @@
           - "wireguard-tools"
         state: present
   when:
-    - (wireguard_rockylinux8_installation_method == "standard" and (ansible_architecture != "x86_64" or ansible_kernel is not match("4\.18\.0-.*x86_64"))) or
-      (wireguard_rockylinux8_installation_method == "dkms")
+    - wireguard_rockylinux8_installation_method == "dkms"
+
--- a/tasks/setup-rocky-8.yml
+++ b/tasks/setup-rocky-8.yml
@@ -47,5 +47,5 @@
           - "wireguard-tools"
         state: present
   when:
-    - wireguard_rockylinux8_installation_method == "non-standard"
+    - wireguard_rockylinux8_installation_method == "dkms"
 
--- a/tasks/setup-rocky-8.yml
+++ b/tasks/setup-rocky-8.yml
@@ -35,6 +35,12 @@
         name: jdoss/wireguard
         chroot: epel-8-{{ ansible_architecture }}
 
+    - name: (Rocky Linux 8) Install EPEL repository
+      ansible.builtin.yum:
+        name:
+          - epel-release
+        update_cache: true
+
     - name: (Rocky Linux 8) Ensure WireGuard KMOD package is removed
       ansible.builtin.yum:
         name:
--- a/tasks/setup-rocky-8.yml
+++ b/tasks/setup-rocky-8.yml
@@ -24,7 +24,8 @@
           - "wireguard-tools"
         state: present
   when:
-    - wireguard_rockylinux8_installation_method == "standard"
+    - (wireguard_rockylinux8_installation_method == "standard" and ansible_architecture == "x86_64" and ansible_kernel is match("4\.18\.0-.*x86_64")) or
+      (wireguard_rockylinux8_installation_method == "kmod")
 
 - name: (Rocky Linux 8) Tasks for non-standard kernel
   block:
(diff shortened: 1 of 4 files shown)


githubixx (commented):
Thanks for the contribution Can you please add an additional test VM in the Molecule test? You can copy https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/molecule.yml#L89-L98 , adjust it accordingly and put it at https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/molecule.yml#L139 . Same for the variables of that new VM which you can copy from https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/molecule.yml#L188-L192 , adjust it and put it at https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/molecule.yml#L214 . That at least enables me to test the change. Thx!

gitouche-sur-osm:
Thank you for walking me through this. Hopefully I got this right.

githubixx:
Thanks for adding the test setup. In general LGTM. But what I'm missing is this "non-standard" thingy :wink: Currently both Rocky8 VMs using the same image and the same kernel. As you might have seen already the [CentOS7 install](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-centos-7.yml#L34) has a task that installs `kernel-plus`. That was actually a bad idea by myself to accept that PR in that form as it's not the responsibility of this role to install kernels.
But nevertheless for testing there is a [prepare.yml](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/prepare.yml) which installs packages or something like that upfront that is not handled by the role itself. Maybe you can just spy a little bit on that [kernel-plus part](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-centos-7.yml#L34-L69) and extend [prepare.yml](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/prepare.yml) accordingly. It doesn't need to work right away. I can adjust it myself in case of problems. But I need a starting point. Testing `aarch64` and `ppc64le` is most probably not easily doable without a lot of effort (if it works at all). But if you use a special kernel that works with a `x86_64/AMD64` VM then it can be added to [prepare.yml](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/kvm/prepare.yml).
Otherwise we most probably will just keep it as is. I guess people who needs that kind of setup know what they're doing :smile:

gitouche-sur-osm:
I understand how confusing it might look.

The "non-standard" comes from the wireguard [installation instructions for Centos8](https://www.wireguard.com/install/#centos-8-module-plus-module-kmod-module-dkms-tools), method 3.

A better name maybe would be **dkms**. As opposed to method 2 (which you have as default method in this ansible role), that uses **kmod**, is very fast and efficient but limited to official kernels on x86_64 architecture.

Here is a summary : 

| RockyLinux 8                     | kmod (elrepo)              | dkms [(copr jdoss)](https://copr.fedorainfracloud.org/coprs/jdoss/wireguard/) |
|----------------------------------|----------------------------|-------------------------------------------------------------------------------|
| x86_64 official kernel           | OK                         | OK                                                                            | 
| x86_64 custom kernel             | will fail to install                 | OK                                                                            |
| arm64 official kernel | no arm64 support in elrepo      | OK             |
| arm64 for RPI [comes with a custom kernel](https://wiki.rockylinux.org/en/special-interest-groups/alt-arch/raspberry-pi) | no arm64 support in elrepo | OK                                                                            |

I see what you mean as testing **dkms** method on **x86_64 official kernel** is not that useful (although it should work just file). I think some different kernel might be possible to install from **elrepo** https://elrepo.org/linux/kernel/el8/x86_64/RPMS/ and will look into it - and test the dkms installation against it.

githubixx:
Don't put too much effort in it. If that "normal" is also ok then we just keep it. I just thought that maybe there needs something more "non-standard" to be installed :wink:

gitouche-sur-osm:
It's fine, I'll look into it when I get the time.  


For now I changed the `non-standard` to `dkms` as the value of `wireguard_rockylinux8_installation_method` variable. It makes more sense. I updated `defaults/main.yml` to reflect the change and provide a better explanation.

tasks/setup-rocky-8.yml
```diff
@@ -2,22 +2,57 @@
...
-      - "wireguard-tools"
-    state: present
+    - name: (Rocky Linux 8) Install WireGuard packages
+      ansible.builtin.yum:
+        name:
+          - "kmod-wireguard"
+          - "wireguard-tools"
+        state: present
+  when:
+    - (wireguard_rockylinux8_installation_method == "standard" and ansible_architecture == "x86_64" and ansible_kernel is match("4\.18\.0-.*x86_64")) or
```
githubixx, line 27: That's way too specific. I don't support any other architecture then `amd64` as I'm not able to test it and therefore not able to maintain it. If there are three use cases like `standard', `kmod` and 'dkms' then there should be three `block`'s. If someone has an `arm64` kernel and wants to use one of the three methods then he/she can do so.

gitouche-sur-osm: Fair enough. I'm reverting back to `standard`/`dkms`as it was.  
This is good enough for me to use. For my use case, I'll find another way to dynamically setup the variable before running the role.

If everything is in order, I'm done editing my branch. Thank you!

githubixx (approved):

Issue #136, opened 2021-12-22
SaveConfig is always true


almereyda:
Even when setting `wireguard_save_config: "false"` explicitly on each host or for the group, all of their configurations will have `SaveConfig = true` present.

githubixx:
Just remove the variable `wireguard_save_config`. Currently the implementation only checks if `wireguard_save_config` is present and if so adds `SaveConfig = true`. Otherwise the setting won't be set. ATM the implementation doesn't care about the variable value. I guess this is something I should change.

PR #149, opened 2022-02-17
honor wireguard_save_config value

githubixx:
ATM the template rendering the WireGuard configuration only checks if `wireguard_save_config` is set and if so sets `SaveConfig = true`. So setting `wireguard_save_config: "false"` had no effect. This PR fixes this behavior.

Fixes https://github.com/githubixx/ansible-role-wireguard/issues/136


almereyda:
Merci

PR #151, opened 2022-03-26
Allow user customize wg.conf file

sckol:
Now etc/wireguard/wg.conf.j2 template is hardcoded into the role. This PR proposes allow user change the template to implement some custom network layout.
Unortunately, it is impossible in Ansible without changing the role, see:
https://stackoverflow.com/questions/55742881/is-there-a-way-to-override-a-template-defined-into-an-ansible-galaxy-role

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -15,6 +15,9 @@ wireguard_port: "51820"
 # The default interface name that WireGuard should use if not specified otherwise.
 wireguard_interface: "wg0"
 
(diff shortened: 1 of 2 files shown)


githubixx:
Thanks for the PR! But I already came to the conclusion that this doesn't makes that much sense. Also see https://github.com/githubixx/ansible-role-wireguard/pull/133 . The whole logic basically is in this template. If someone needs a different logic then I think it's better to fork this role.

PR #152, opened 2022-03-28
Introduce config switch for specifying the interface restart behavior

### LINKED ISSUE #147: Routes not updating during re-configuration
When an existing Wireguard interface gets updated, on newer system [wg syncconf gets used](https://github.com/githubixx/ansible-role-wireguard/blob/c4a5677f723fb8ea3224e5753b9433498725d42f/handlers/main.yml).

The issue with `wg syncconf` seems to be that it [does not update routes](https://www.reddit.com/r/WireGuard/comments/s5cji1/wg_setconf_synccong_and_ip_routes/). In our case, we modified the routes of an existing Wireguard link, which led to the case that the wireguard handshakes were successfully sent and received, but we weren't able to transmit any data due to the missing routes.

Restarting the interfaces manually, solved the issue.

Maybe it would be nice to document this behavior and use the "old" restart mechanism by default as this issue can be quite hard to detect and debug.

lmm-git:
This change was made in order to enable the possibility to ensure e.g. routes of the Wireguard interface get properly reloaded when they change while sacrificing uninterrupted interfaces.

I still would argue that most users would want the properly reloaded interfaces instead of an fully uninterrupted interface by default as when having e.g. no proper routes set it is quite hard to detect. Even more, simply restarting the server / interface does fix the issue, which makes it even harder to debug if users are not too familiar with how Wireguard configuration works.

I know that changing the default behavior would be a breaking change but I believe this change would not affect / be noticeable for many users. If you agree I will change the PR so the default is `restart` instead of auto-detect.

Fixes #147

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,6 +75,14 @@ wireguard_conf_group: "{{ 'root' if not ansible_os_family == 'Darwin' else 'whee
 # The default mode of the wg.conf file
 wireguard_conf_mode: 0600
 
+# Which method should be used to apply new wireguard configurations?
+# Possible options:
+#   sync: Does make use of the `wg syncconf` command, connections are not interrupted but it does
+#         populate route changes to the interface. Not available on older kernels / systems.
+#   restart: Performs a full restart on the interface. Does interrupt the connections shortly.
+#   undefined: Use sync on systems where it is available, otherwise use restart. (default)
+# wireguard_interface_restart_behavior:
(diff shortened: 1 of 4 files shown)


githubixx:
I've implemented it a little bit differently in https://github.com/githubixx/ansible-role-wireguard/pull/156. So closing this one.

PR #154, opened 2022-04-30
fix: Do not require lsb-release on Debian

blackandred:
Same as in https://github.com/DataDog/ansible-datadog/pull/377


githubixx (approved):
Thanks!

PR #158, opened 2022-05-13
Add a initial support of Windows

yannickt:
As a first contribution, I hope I did a not-to-bad job.


githubixx (changes requested):
Thanks for your contribution! In general looks good so far. I've added a few comments already and maybe I'll add a few more later. The problem with Windows is of course that it's not easily testable which isn't that nice. Bat as long as it doesn't interfere with the existing code it's most probably acceptable.
Please also add an entry to `CHANGELOG.md`. The new version will be `9.3.0`.

PR #161, opened 2022-09-13
add by host custom allowed ips

vincentDcmps:
hello 
I have apply change to customize allowed_ips in function of host.

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,9 @@ You can specify further optional settings (they don't have a default and won't b
 
 ```yaml
 wireguard_allowed_ips: ""
+wireguard_byhost_allowed_ips:
+  host1: 10.0.0.2,192.168.1.41,192.168.1.0/24
+
 wireguard_endpoint: "host1.domain.tld"
 wireguard_persistent_keepalive: "30"
 wireguard_dns: "1.1.1.1"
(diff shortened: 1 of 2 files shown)


githubixx:
Hello! I'm not so sure for what this change might be needed TBH :wink: You can set `wireguard_allowed_ips` per host already and if you want host routes you can assign a value of `10.0.0.2/32,192.168.1.41/32` e.g.

vincentDcmps:
in my case I have three device with a central device 

with my modification by example on gerard and oscar:
```
wireguard_byhost_allowed_ips:
  merlin: 10.0.0.6,192.168.1.41
```

```mermaid
flowchart LR
  A[oscar] <--> B[merlin]
  C[gerard] <--> B 
  
```
so I don't want that a and communicate directly by wireguard themself because they are on same lan

if I set `wireguard_allowed_ips`  like you say I will have more something like that

```mermaid
flowchart LR
  A[oscar] <--> B[merlin]
  C[gerard] <--> B 
  A <--> C
```

githubixx:
I somehow still don't get this PR :wink: Personally it seems wrong to me to have a "global" variable where you define a dictionary where the hostname is the key while you have the Ansible's host repository on the other side. So if you have something specific that only applies to one host why not use host variables? :confused: 

I guess this Molecule test comes more or less close to your use case: https://github.com/githubixx/ansible-role-wireguard/tree/master/molecule/kvm-single-server Can you maybe use that one as a template and adjust it accordingly so that it matches your use case? You don't need to execute it as you most probably don't have Vagrant and KVM. But it'd give me an idea.

Unit193:
@vincentDcmps I'm a little late to the party, but thanks for filing this PR.  It's exactly what I needed! :)

PR #163, opened 2022-10-09
add rocky 9 suport

vincentDcmps:
hello, I have add support for rocky 9 for this role test with molecul output is sucess


trunet:
I'm writing the same for AlmaLinux 9 and I think it's exactly the same as Rocky Linux 9 (both based on RHEL 9)

These newer OSs don't need kmod-wireguard (and it's not provided by elrepo).

`jdoss/wireguard` copr doesn't have any builds for AlmaLinux 9, can you confirm on Rocky Linux? If the same, it also doesn't support dkms.

Having said that, I think the only way to install in Alma/Rocky is straight installing wireguard-tools, as wireguard support is embedded into the kernel and no extra repositories are needed.

vincentDcmps:
I following your recomandation I merge alma and rocky setups

githubixx:
Thanks for the PR! I've implemented a little bit different solution at the end and created https://github.com/githubixx/ansible-role-wireguard/pull/168

PR #164, opened 2022-10-24
Add support to AlmaLinux 9

trunet:
This adds support to AlmaLinux 9.

I'm using this fork myself and it's working for me. I have a mac, therefore I couldn't run molecule tests. I appreciate if anybody could run the tests and let me know the results.

### DIFF
--- a/meta/main.yml
+++ b/meta/main.yml
@@ -24,6 +24,7 @@ galaxy_info:
       versions:
         - "7"
         - "8"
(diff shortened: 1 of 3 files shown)


githubixx:
Thanks for the PR! I've implemented a little bit different solution at the end and created https://github.com/githubixx/ansible-role-wireguard/pull/168

Issue #126, opened 2021-10-01
Ubuntu (and systemd) DNS specifics


gdanov:
Executive summary: when the DNS option is set, wg-quick does not run on Ubuntu without openresolv. Installing (and using) openresolv, messes up with the standard way resolution is handled (via systemd-resolved). Use systemd-resolve on such distros to properly inject WG DNS servers.

I'm working around this by keeping the DNS empty and using the following command in PostUp:
`systemd-resolve -i wg0 --set-dns=10.99.1.10 --set-dns=10.99.1.11 --set-dns=10.99.1.12`

Would be nice if the role detected this and implemented the workaround automatically.

Alternative is to use systemd's netdevs, but that's pretty heavy-handed change.

githubixx:
Not sure if I can do that much in this case. According to https://systemd.io/RESOLVED-VPNS/ (`The APIs` -> `3`): 

```
If you don’t want to use any systemd-resolved commands,
you may use the resolvconf wrapper we provide. resolvectl
is actually a multi-call binary and may be symlinked to resolvconf,
and when invoked like that behaves in a way that is largely
compatible with FreeBSD’s and Ubuntu’s/Debian’s resolvconf(8) tool. 
```
I just tried that out on Ubuntu 20.04:
```
host: > /usr/bin$ sudo ln -s resolvectl resolvconf
host: > /usr/bin$ resolvconf --help
resolvconf -a INTERFACE < FILE
resolvconf -d INTERFACE

Register DNS server and domain configuration with systemd-resolved.

  -h --help     Show this help
     --version  Show package version
  -a            Register per-interface DNS server and domain data
  -d            Unregister per-interface DNS server and domain data
  -f            Ignore if specified interface does not exist
  -x            Send DNS traffic preferably over this interface

This is a compatibility alias for the resolvectl(1) tool, providing native
command line compatibility with the resolvconf(8) tool of various Linux
distributions and BSD systems. Some options supported by other implementations
are not supported and are ignored: -m, -p. Various options supported by other
implementations are not supported and will cause the invocation to fail: -u,
-I, -i, -l, -R, -r, -v, -V, --enable-updates, --disable-updates,
--updates-are-enabled.

See the resolvectl(1) man page for details.
```
[...]

PR #169, opened 2022-11-19
install Systemd resolved in rocky

vincentDcmps:
when use 'wireguard_dns' variable in rocky 9 you get an error on wireguard start because resovconf not exist.
I propose to install systemd-resolved has a workaround only when 'wireguard_dns' variable is defined

### DIFF
--- a/tasks/setup-rocky.yml
+++ b/tasks/setup-rocky.yml
@@ -6,3 +6,16 @@
   ansible.builtin.yum:
     name: wireguard-tools
     state: present
+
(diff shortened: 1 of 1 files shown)


githubixx:
We already had a similar discussion at https://github.com/githubixx/ansible-role-wireguard/issues/126 but for Ubuntu. As mentioned there DNS requirements should be installed and setup outside of this role as everybody has different requirements in this regards. Nevertheless thanks for the pull request!

PR #178, opened 2023-01-31
Implementation of pre shared keys per peer

packetplane:
With a quick search a found this PR #58 that was having a static preshared key for all peers. Preferably your peers all have unique preshared keys, so this is what is implemented in this pull request.

This is a draft implementation of preshared keys per peer.

Newly added peers / hosts in your inventory will automatically be set up with Preshared Keys. 
Existing peers won't be updated.

I would love to write some automated tests for this. I'm not quite familiar with molecule just yet. So that's a work in progress.

### DIFF
--- a/templates/etc/wireguard/wg.conf.j2
+++ b/templates/etc/wireguard/wg.conf.j2
@@ -51,6 +51,8 @@ SaveConfig = {{ wireguard_save_config }}
 # {{ host }}
 PublicKey = {{hostvars[host].wireguard__fact_public_key}}
 {%     if  (
+         hostvars[host].wireguard__internal_preshared_keys is defined and
+         wireguard__internal_preshared_keys is defined and
          inventory_hostname in hostvars[host].wireguard__internal_preshared_keys and
          host in wireguard__internal_preshared_keys
        )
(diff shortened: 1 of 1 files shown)


githubixx:
@daanh432 Are you still working on this? Otherwise I think it'd make sense to close it for now.

daanh432:
I have been running the pre shared keys patch for a while now but sadly I did not have the time yet to look into Molecule and writing automated tests. The earliest moment I might have some spare time to look into it would be in July.

Unit193:
Thanks for working on this!  This is one of the blockers preventing me from using this role.

In a test setup with preexisting config, when I run the role (with only a few of the endpoints defined), I get the following error:

fatal: [vanir]: FAILED! => {"msg": "The task includes an option with an undefined variable. The error was: list object has no element 5. list object has no element 5\n\nThe error appears to be in '/home/unit193/.ansible/roles/githubixx.ansible_role_wireguard/tasks/main.yml': line 175, column 7, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n    - name: Set pre shared key fact\n      ^ here\n"}

So it would seem having hosts in the target system's config with PSKs that are not yet in ansible config causes this to bail?

daanh432:
I'm not invested in an Ansible wireguard setup anymore. If someone wants to pick this up in the future feel free to do so. 
I will close this PR as I won't be continuing on this myself.

PR #179, opened 2023-02-22
option wireguard_update_cache to disable refresh

sebix:
previously disable the cache update was only possible on ubuntu using the generic name wireguard_update_cache it can be enabled for all operating systems

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,9 @@ wireguard_interface_restart: false
 # course a very sensitive value you might consider a tool like Ansible Vault
 # to store it encrypted.
 # wireguard_private_key:
+
+# Set to "false" if package cache should not be updated
+wireguard_update_cache: "true"
 ```
 
 There are also a few Linux distribution specific settings:
@@ -148,8 +151,9 @@ There are also a few Linux distribution specific settings:
 # - elementary OS
 #######################################
 
-# Set to "false" if package cache should not be updated
-wireguard_ubuntu_update_cache: "true"
+# Set to "false" if package cache should not be updated.
+# DEPRECATED: This variable will be removed with the next major release. Please use "wireguard_update_cache" instead.
+wireguard_ubuntu_update_cache: "{{ wireguard_update_cache }}"
 
 # Set package cache valid time
 wireguard_ubuntu_cache_valid_time: "3600"
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -74,7 +74,9 @@ wireguard_interface_restart: false
 #######################################
 
 # Set to "false" if package cache should not be updated
-wireguard_ubuntu_update_cache: "true"
+wireguard_update_cache: "true"
+# DEPRECATED: This variable will be removed with the next major release. Please use "wireguard_update_cache" instead.
+wireguard_ubuntu_update_cache: "{{ wireguard_update_cache }}"
 
 # Set package cache valid time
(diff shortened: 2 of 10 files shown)


molecule/kvm-single-server/prepare.yml
```diff
@@ -9,5 +9,5 @@
   tasks:
     - name: Update APT package cache
       ansible.builtin.apt:
-        update_cache: true
+        update_cache: "{{ wireguard_update_cache }}"
```
githubixx, line 12: I actually what to keep it as is. This a Molecule test and in this phase only the VMs and a little bit of other stuff gets initialized. So at this phase the Wireguard role isn't involved.

molecule/kvm/prepare.yml
```diff
@@ -43,7 +43,7 @@
   tasks:
     - name: Update APT package cache
       ansible.builtin.apt:
-        update_cache: true
+        update_cache: "{{ wireguard_update_cache }}"
```
githubixx, line 46: Same as above.

README.md
```diff
@@ -150,9 +153,6 @@ There are also a few Linux distribution specific settings:
 # - elementary OS
 #######################################
 
-# Set to "false" if package cache should not be updated
-wireguard_ubuntu_update_cache: "true"
```
githubixx, line 154: I actually want to keep that variable for backward compatibility at least for now. So this could be a possible solution:
```
# Set to "false" if package cache should not be updated.
# DEPRECATED: This variable will be removed with the next major release. Please use "wireguard_update_cache" instead.
wireguard_ubuntu_update_cache: "{{ wireguard_update_cache }}"
```

defaults/main.yml
```diff
@@ -74,7 +74,7 @@ wireguard_interface_restart: false
 #######################################
 
 # Set to "false" if package cache should not be updated
-wireguard_ubuntu_update_cache: "true"
+wireguard_update_cache: "true"
```
githubixx, line 77: See my README comment about this.

tasks/setup-elementary os.yml
```diff
@@ -4,7 +4,7 @@
 
 - name: (elementary OS) Update APT package cache
   ansible.builtin.apt:
-    update_cache: "{{ wireguard_ubuntu_update_cache }}"
```
githubixx, line 7: As mentioned above I'd like to keep that variable for now.

tasks/setup-ubuntu.yml
```diff
@@ -4,7 +4,7 @@
 
 - name: (Ubuntu) Update APT package cache
   ansible.builtin.apt:
-    update_cache: "{{ wireguard_ubuntu_update_cache }}"
```
githubixx, line 7: As mentioned above I'd like to keep that variable for now.

githubixx (changes requested):
Thanks for the PR! Makes sense :smiley: I've added a few comments. Also please add a line about that change to CHANGELOG.md. The version should be `13.1.0`. And you might need to rebase as I just released version `13.0.0`.

defaults/main.yml
sebix: `wireguard_ubuntu_update_cache` now defaults to `wireguard_update_cache`

sebix:
Thank you for your feedback!
I finally took the time to implement your suggestions, the PR is now ready for re-review.

Thank you for maintaining this role!

PR #184, opened 2023-03-08
Add reboot to the standard mode

mofelee:
It needs to be rebooted to make the WireGuard kernel module available.

### DIFF
--- a/tasks/setup-centos-7.yml
+++ b/tasks/setup-centos-7.yml
@@ -24,6 +24,13 @@
           - "kmod-wireguard"
           - "wireguard-tools"
         state: present
+      register: wireguard__centos7_yum_updates
+
+    - name: (CentOS 7) Reboot Instance to update kernel
+      ansible.builtin.reboot:
+        reboot_timeout: "{{ wireguard_centos7_standard_reboot_timeout }}"
+      when:
+        - wireguard__centos7_yum_updates.changed is true
 
 - name: (CentOS 7) Ensure WireGuard DKMS package is removed
   ansible.builtin.yum:
@@ -58,11 +65,11 @@
           - "kernel-plus"
           - "wireguard-tools"
         state: present
-      register: centos7_yum_updates
+      register: wireguard__centos7_yum_updates
 
     - name: (CentOS 7) Reboot Instance to update kernel
       ansible.builtin.reboot:
         reboot_timeout: "{{ wireguard_centos7_kernel_plus_reboot_timeout }}"
       when:
-        - centos7_yum_updates.changes is defined
-        - centos7_yum_updates.changes.installed|flatten|select('regex', '^kernel-plus$') is any
+        - wireguard__centos7_yum_updates.changes is defined
+        - wireguard__centos7_yum_updates.changes.installed|flatten|select('regex', '^kernel-plus$') is any
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@ SPDX-License-Identifier: GPL-3.0-or-later
 
 # Changelog
 
+## 14.0.0
+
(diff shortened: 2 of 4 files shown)


tasks/setup-centos-7.yml
```diff
@@ -24,6 +24,13 @@
           - "kmod-wireguard"
           - "wireguard-tools"
         state: present
+      register: centos7_yum_updates
```
githubixx, line 27:
```suggestion
      register: wireguard__centos7_yum_updates
```
Internal variables should be prefixed with `wireguard__`.

As you're already on it :wink: : Could you please also add this prefix here:

https://github.com/githubixx/ansible-role-wireguard/blob/17ed7be8f324a9521193e193df3b7d2bb5c7abc5/tasks/setup-centos-7.yml#L68
https://github.com/githubixx/ansible-role-wireguard/blob/17ed7be8f324a9521193e193df3b7d2bb5c7abc5/tasks/setup-centos-7.yml#L74-L75

That was an oversight to not add it. Thanks!

tasks/setup-centos-7.yml
```diff
@@ -24,6 +24,13 @@
...
+
+    - name: (CentOS 7) Reboot Instance to update kernel
+      ansible.builtin.reboot:
+        reboot_timeout: "{{ wireguard_centos7_standard_reboot_timeout }}"
+      when:
+        - centos7_yum_updates.changed is true
```
githubixx, line 33:
```suggestion
        - wireguard__centos7_yum_updates.changed is true
```

githubixx (changes requested):
Thanks! A few comments inline. Please also update `README.md` accordingly. Also please add an entry to `CHANGELOG.md` about what was changed. As this actually changes the behavior of for the CentOS 7 users it makes sense to have a new major version number. So it would be `14.0.0`.

mofelee:
@githubixx 
I think it's okay now.

mofelee:
In my case, if I don't reboot CentOS 7, it cannot pass through the 'Enable WireGuard kernel module' task.

```bash
> cat /etc/os-release 
NAME="CentOS Linux"
VERSION="7 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="7"
PRETTY_NAME="CentOS Linux 7 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:7"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

CENTOS_MANTISBT_PROJECT="CentOS-7"
CENTOS_MANTISBT_PROJECT_VERSION="7"
REDHAT_SUPPORT_PRODUCT="centos"
REDHAT_SUPPORT_PRODUCT_VERSION="7"
```

```bash
> uname -a
Linux xx.local 3.10.0-957.1.3.el7.x86_64 #1 SMP Thu Nov 29 14:49:43 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

```bash
> cat /etc/centos-release
CentOS Linux release 7.6.1810 (Core)
```

```bash
> lsb_release
LSB Version:	:core-4.1-amd64:core-4.1-noarch
```


CleanShot 2023-03-17 at 09 24 44@2x


CleanShot 2023-03-17 at 09 27 10@2x

mofelee:
CleanShot 2023-03-17 at 09 54 24@2x

I've reproduced this error on this ISO file.

http://bay.uchicago.edu/centos-vault/7.6.1810/isos/x86_64/CentOS-7-x86_64-Minimal-1810.iso

http://bay.uchicago.edu/centos-vault/7.6.1810/isos/x86_64/

githubixx (approved):

PR #188, opened 2023-05-27
Make it possible to use wg-config tag only

donat-b:
Currently it would fail with "'wireguard__restart_interface' is undefined"


githubixx (approved):
Thanks!

PR #189, opened 2023-06-27
Add support for RHEL 8

Matthew-Beckett:
This PR adds support for executing the playbook on RHEL 8 based systems. It does not use DKMS and is based on the Alma Linux 8 implementation.

This has been tested on RHEL 8.7.

### DIFF
--- a/tasks/setup-redhat-8.yml
+++ b/tasks/setup-redhat-8.yml
@@ -1,0 +1,20 @@
+---
+- name: (RedHat 8) Install EPEL & ELRepo repository
+  ansible.builtin.yum:
+    name:
+      - epel-release
+      - elrepo-release
+    update_cache: "{{ wireguard_update_cache }}"
+
+- name: (RedHat 8) Ensure WireGuard DKMS package is removed
+  ansible.builtin.yum:
+    name:
+      - "wireguard-dkms"
+    state: absent
+
+- name: (RedHat 8) Install WireGuard packages
+  ansible.builtin.yum:
+    name:
+      - "kmod-wireguard"
+      - "wireguard-tools"
+    state: present
--- a/tasks/setup-redhat-8.yml
+++ b/tasks/setup-redhat-8.yml
@@ -2,8 +2,8 @@
 - name: (RedHat 8) Install EPEL & ELRepo repository
   ansible.builtin.yum:
     name:
-      - epel-release
-      - elrepo-release
+      - https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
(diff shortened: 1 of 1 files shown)


githubixx (commented):
Thanks for the PR! This is pretty similar to what is already implemented for [Rocky Linux 8]https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-rocky-8.yml#L5-L27] and for [Alma Linux 8](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-almalinux-8.yml)

The problem with Red Hat Enterprise is that it's not possible to have a Molecule test as you can't get a copy for free as a private person and there is also no KVM image for that. And now it got even worse that they decided to go closed source. So it's basically not possible test Red Hat 8 directly (only the derivatives) with my Molecule tests and that's not really nice. So merging this would be like "might work, but :shrug: " :wink: 

Maybe it makes sense to unify this PR and the non DKMS code for Alma Linux 8 and Rocky Linux 8 as they should be basically the same if I'm not completely wrong. That would allow to run the tests on Rocky/Alma 8 at least as they share the same code base and the probability is most probably pretty high then that it will also work with Red Hat 8. As it's currently implemented this PR wouldn't have any tests.

Matthew-Beckett:
> Thanks for the PR! This is pretty similar to what is already implemented for [Rocky Linux 8][https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-rocky-8.yml#L5-L27]](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-rocky-8.yml#L5-L27%5D) and for [Alma Linux 8](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/setup-almalinux-8.yml)
> 
> The problem with Red Hat Enterprise is that it's not possible to have a Molecule test as you can't get a copy for free as a private person and there is also no KVM image for that. And now it got even worse that they decided to go closed source. So it's basically not possible test Red Hat 8 directly (only the derivatives) with my Molecule tests and that's not really nice. So merging this would be like "might work, but shrug " wink
> 
> Maybe it makes sense to unify this PR and the non DKMS code for Alma Linux 8 and Rocky Linux 8 as they should be basically the same if I'm not completely wrong. That would allow to run the tests on Rocky/Alma 8 at least as they share the same code base and the probability is most probably pretty high then that it will also work with Red Hat 8. As it's currently implemented this PR wouldn't have any tests.

I agree, makes sense. I will work to unify this into a single RHEL-derived OS role which uses non-dkms installation methods.

PR #193, opened 2023-10-13
feat: check if all hosts are accessible

spyesx:
Check if executing tasks as  root is permitted.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -5,6 +5,12 @@
 - name: Gather instance facts
   setup:
 
+- name: Test SSH and root access
+  become: true
+  raw: "{{ 'true' if ansible_os_family == 'Darwin' else '/bin/true'  }}"
+  changed_when: false
+  any_errors_fatal: true
+
 - include_tasks:
     file: "{{ item }}"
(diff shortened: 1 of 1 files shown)


githubixx (commented):
I'm actually not sure for what this is good for :wink: If a host is not available then it most probably should be removed from Ansible's hosts list. But in general I guess it makes sense to have this kind of functionality outside of this role in a separate playbook and include this role via `ansible.builtin.include_role` into that playbook.

spyesx:
I understand what you mean while using Ansible in a production environment. We also utilize Ansible for development servers and workstations, and Wireguard facilitates access to a backend network for a range of automated administrative tasks.

To ensure seamless operation, we preliminary check to verify that all anticipated hosts are reachable. That simple measure serves two purposes:

1. It allows us to identify and address any connectivity issues early on, preventing potential delays or disruptions.
2. It prevents inadvertent creation of configurations that include missing hosts, which could lead to surprises and complications down the road.

spyesx:
Thx for `ansible.builtin.include_role`. I'll have a look at it too.

PR #194, opened 2023-10-22
Update a link to the role at Ansible Galaxy

tired-engineer:


githubixx:
Thanks, but Ansible Galaxy is currently (well actually since two weeks) completely broken if it comes to Ansible roles. See all the posts here: https://forum.ansible.com/tag/galaxy-ng
I'm currently thinking about removing everything from Ansible Galaxy. It looks like that Ansible roles are not a priority anymore. I'll wait another week and if there is no progress I most probably remove my Galaxy content.

tired-engineer:
Thanks for the explanation, never mind then!

PR #196, opened 2023-11-13
Add support for wireguard_include_peers variable

### LINKED ISSUE #195: ability to restrict exported peers
(this is related to #101, but not exactly the same)

I have a complex wireguard setup where not all nodes can directly talk to each. This works fine at the moment because there is one central node that all peers can reach (and which has an allowed_ips configured that is wide enough to cover all nodes). I only configure peers that are actually reachable.

It would be great if there was a way of explicitly specifying the peers that should be exported for a particular node, and otherwise exporting all available peers.

jelmer:
This allows explicitly setting which peers a node can connect to. If the variable is not present, then all peers are included.

Fixes #195

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,6 +336,13 @@ wireguard_unmanaged_peers:
     persistent_keepalive: 0
 ```
 
+If present, the ``wireguard_include_peers`` setting restricts which peers (as hostnames) are included for a particular host. This can be useful when not all nodes can reach each other:
+
+```yaml
+wireguard_include_peers:
+  - client.example.com
+```
+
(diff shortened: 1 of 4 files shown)


githubixx (commented):
Thanks for your PR! I'm currently not so happy with the variable name `wireguard_reachable_peers` :wink: While the term is "technically" true I think it takes a bit to figure out what this variable is all about. E.g. if the variable would be called `wireguard_include_peers` then it'd be possible to also introduce `wireguard_exclude_peers` later if needed. Formerly that would have been a "whitelist" and "blacklist" I guess but makes no sense anymore to use them.

Nevertheless there is definitely some documentation needed and the variable should be put separately in README. The variables in that variable block are covered by "wg_quick". That's why `wireguard_unmanaged_peers` shouldn't be there too. I'll change that with the next release.

So can you please put your variable and one or two sentences below the text `The commands are executed in order as described in wg-quick.8`?

jelmer:
That seems reasonable - updated to address your feedback.

jelmer:
@githubixx any more thoughts on this?

githubixx:
@jelmer Sorry, I'm currently a little bit short on time. I'll get back to you next week. I need to test the patch first to make sure it has no side effects. But it'd be helpful if you could extend the [Molecule test](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/default/molecule.yml) accordingly. You don't have to execute it. It's just for me to have a starting point. Maybe make a copy of the [Ubuntu 22.04](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/default/molecule.yml#L150-L161) host and put it [here](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/default/molecule.yml#L222) with adjusted values. And the add the needed variable at the end [here](https://github.com/githubixx/ansible-role-wireguard/blob/master/molecule/default/molecule.yml#L326). Thanks!

jelmer:
updated to resolve conflicts; I hope to get to adding tests sometime this month

kbcz1989:
@jelmer Hello. Thanks for getting back at this. I am in dire need of this change :)
I wanted to make tests to get this going so Ill paste my change here, if it helps you anyhow:
```diff
-    - name: There should be as much WireGuard interfaces as hosts in vpn group minus one
+    - name: There should be as many WireGuard interfaces as hosts in vpn group minus one or as defined peers
       ansible.builtin.assert:
         that:
-          - "hosts_count|int -1 == wireguard__interfaces_count.stdout|int"
+          - >
+            (wireguard_include_peers is defined and
+             wireguard__interfaces_count.stdout|int == wireguard_include_peers|length) or
+            (wireguard_include_peers is not defined and
+             hosts_count|int - 1 == wireguard__interfaces_count.stdout|int)
```

jelmer:
@githubixx I've added a molecule test

PR #200, opened 2024-01-04
Enable skipping of wg kernel module install

### LINKED ISSUE #190: Proxmox LXC containers attempting to insert kernel modules
I've got a set of unprivileged Debian 11 LXCs running in Proxmox which are attempting to insert the WireGuard kernel module when running this role.  Looking at the logic [here](https://github.com/githubixx/ansible-role-wireguard/blob/master/tasks/main.yml#L34), I'm wondering if there needs to be an additional flag set in the PVE guest bits which will turn off the insert?  Normally I'd suggest checking to see if the module is already loaded, but inside the LXC we can't.

To be clear, completely disabling the load makes things work just fine since the host already has the relevant module loaded.

Edit: This seems similar to #103, but maybe from the other side of the coin?

gregorydlogan:
This PR enables skipping the wg kernel module install based on tag (`wg-install-kernel-module`) or role variable `wireguard_install_kernel_module`.

Fixes #190

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,12 @@ wireguard_interface_restart: false
 # Set to "false" if package cache should not be updated (only relevant if
 # the package manager in question supports this option)
 wireguard_update_cache: "true"
+
+# Normally the role installs and activates the wireguard kernel module where
+# appropriate.  In some cases we might not be able load kernel modules, like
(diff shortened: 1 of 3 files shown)


githubixx (approved):
LGTM. Thanks!

PR #201, opened 2024-02-26
Wireguard conf backup

shk3bq4d:
Setting up tunnels can be error prone and it's nice to have the ability to review a past deployed working configuration when one accidently breaks a tunnel.

By simply leveraging the `backup` directive of the `template` ansible module, one easily let the role user control if that mechanism is useful to them.

Best regards


githubixx (approved):
LGTM. Thanks!

PR #205, opened 2024-06-07
Update setup-fedora.yml

ahjohannessen:
Fedora CoreOS has full support for WireGuard out of the box and does not have yum.

This change makes it possible to setup wg on FedoraOS.


ahjohannessen:
@githubixx WDYT?

githubixx (approved):
Thanks for the PR! Sadly there is no Vagrant image for libvirt to be able to test it. But the change seems to be simple enough to include it with the next release if it makes this OS work. But I wont list it as a "supported" OS because it can't be tested.

ahjohannessen:
I understand :) Thanks for the merge

PR #210, opened 2024-10-14
Netplan support

kbcz1989:
Hello,

first of all thank you for creating such a great role. This PR is adding support for managing wireguard through [Netplan](https://netplan.io/).
It's non-invasive change, nothing should happen if variable wireguard_use_netplan stays false, as defined in defaults.

Thank you!


githubixx (commented):
Thanks for the PR! I'll try to have a look till end of next week or so. But one thing upfront: Netplan is specific to Ubuntu, right? In this case it should be `wireguard_ubuntu_use_netplan` and moved to the section were the other `wireguard_ubuntu_*` variables are defined.

kbcz1989:
@githubixx Thank you for the review and suggestions. I updated the code to reflect it.

githubixx (changes requested):

kbcz1989:
@githubixx Great suggestions, thank you! I updated the code.

Molecule tests are executed successfully:
```shell
PLAY RECAP *********************************************************************
test-wg-ubuntu2004         : ok=23   changed=4    unreachable=0    failed=0    skipped=7    rescued=0    ignored=0
test-wg-ubuntu2204         : ok=23   changed=4    unreachable=0    failed=0    skipped=7    rescued=0    ignored=0
test-wg-ubuntu2404         : ok=23   changed=4    unreachable=0    failed=0    skipped=7    rescued=0    ignored=0
```

I also updated the docs, let me know if I provided enough info or if there is anything else I should change

kbcz1989:
@githubixx I implemented changes made in #102, but I had to update the logic as changes in that PR breaks existing setups because if wireguard_endpoint is undefined (which is default) the peers and port will not be set.

Please let me know if I should open a separate PR for handling both netplan template as well as the original template or how should this be handled?

kbcz1989:
@githubixx I now realize that molecule tests needs to be properly updated also. I will look into it.

kbcz1989:
@githubixx The tests you prepared are actually covering netplan use-case correctly. I think the PR is now ready for review. Can you please check?

Once and if this gets merged I will look into 2 more issues:
1. fix the conditions introduced in fa7b0db and make proper molecule tests for that
2. implement changes from #196 and make proper tests for that also

githubixx (commented):
I only found three minor issues left. If they're fixed I guess the PR can be merged.

kbcz1989:
@githubixx thanks, it's done now.

